PR #6, opened 2020-08-14
Proposal classes

ingdanielguerrero:
There are three classes:

- **sepalwidgets**: proposal to create sepal widgets based on ipyvuetify with some useful behaviors. With that in mind we could create templates for all the derived elements.
- **aoi**: proposal aoi class to manage the asset components, intended to put together some behaviors related with an aoi element, such as the existent in other modules: drawing shapes, upload shapefiles, and so on.
- mapping: class inheriting directly from geemap, intended to create a unique map to be used among other apps.

### DIFF
--- a/sepal_ui/aoi.py
+++ b/sepal_ui/aoi.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python3
+
+from . import mapping
+
+import ee
+
+
+
+class Aoi_IO:
+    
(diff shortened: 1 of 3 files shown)


12rambau:
I refused the commit into the master branch because some product already in use are fully based on the old implementation. 
I added your changes to the develop branches and I'll start a review tomorrow

PR #36, opened 2020-10-20
Alert methods

ingdanielguerrero:
- Append method to alert widget.
- Added type_ trait to Divider class, so default colors can be changed by specifying the type_ parameter.
- Commented examples of usage.


12rambau:
you will need to add the test for your new function when I'll push the test branch in master

PR #77, opened 2020-12-08
Vue templates

ingdanielguerrero:
This new branch has a new `templates` directory to store those templates which are useful to mimic the Vuetify implementations, as first implementation I created a `DatePickert` object to test it, while we can decide which path choose, or until there is a solution for the Bug in Ipyvuetify [https://github.com/mariobuikhuizen/ipyvuetify/issues/88](url).
What do you think? I've created an example notebook to test it.

### DIFF
--- a/examples/Alert.ipynb
+++ b/examples/Alert.ipynb
@@ -13,27 +13,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import sys\n",
-    "sys.path.append('..')"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
     "import sepal_ui.sepalwidgets as sw"
-   ]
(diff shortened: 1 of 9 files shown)


12rambau:
Can you move the pull request to develop ? 
The idea is the following : 
- we push our idea on develop 
- we add the appropriate tests in develop
- we push develop to master 
- the test evaluation is performed on master (I cannot change this behaviour)

12rambau:
I can't test you implementation because the templates folder is not available in the repository (I think `templates`is a keyword in the `.gitignore`)

12rambau:
As aggreed we will pull to develop instead of going for maser (to keep the test under control and allow ourself to make small mistakes)

ingdanielguerrero:
yes sure, I'm gonna see what's happening with the templates folder, and why is not being installed along with the rest of the branch, then I'll merge it with develop for your review.

PR #166, opened 2021-03-16
Fix the add_raster method

### LINKED ISSUE #135: the add_raster function does'nt work since February in SEPAL
The wheel keep spinning and nothing is loaded on the map

12rambau:
we also need to wait for the merge of https://github.com/davidbrochart/xarray_leaflet/pull/41 to make it work in JupyterLab but the voila and notebook version should be repaired. 

Fix #135


ingdanielguerrero (approved):
Perfect!

PR #244, opened 2021-08-06
use the black code formatter 

### LINKED ISSUE #231: use the PEP 8 convention on code length
https://www.python.org/dev/peps/pep-0008/#maximum-line-length as specified in PEP 8 we should limit ourselves to 80 characters: 

- [ ] refactor the source code 
- [ ] add a test to check it

### LINKED ISSUE #238: use the black lib to rationalize the code structure
We cannot agree amongs the developers on a way to write code: stop at 80, stop at 120, never stop.... 

Instead of spending countless hours trying to find a politically accepted answer, @dfguerrerom proposed to use a lib to decide for us : https://black.readthedocs.io/en/latest/

black is one of the leading python code formatter, if Django uses it, I see no reason not to use it as well.

12rambau:
Fix #238
Fix #231 

This modification includes: 
- add `black` to the requirements 
- apply `black` on the whole lib 
- add `black` in the CI tests 
- add `black` as a pre-commit hook (the first launch takes time the folowings takes seconds)


dfguerrerom (approved):
61 file reviewed ;)

Issue #253, opened 2021-08-31
Validate selected and add default assets to Asset Select widget


dfguerrerom:
It's nice to have a default asset for the user but it would be also nice to have a list of custom/predefined assets, so the user will have the option to select one or another and keep them in the list without losing them.
There is a missing validation of the inputs, would be nice to have an alert message once an asset is selected and store the status in the widget attributes, such as `valid` and `type`.

dfguerrerom:
Instead of having `asset type`, I think that it would be more useful to store the full `asset_info`, so it could be used afterward.

12rambau:
> Instead of having asset type, I think that it would be more useful to store the full asset_info, so it could be used afterward. What do you think @12rambau ?

agreed

12rambau:
can we close the issue ?

PR #254, opened 2021-08-31
Develop

dfguerrerom:
Intended to solve #253


dfguerrerom:
I forgot to mention, but this will fail on the test due to the change of the `default_asset` type, now it will require a list not a string. What do you think?

dfguerrerom:
- I used an automatic translation to french, so I've zero clues if that makes sense ;).
- I've changed the asset type for asset info, would be more useful.

12rambau:
> I forgot to mention, but this will fail on the test due to the change of the `default_asset` type, now it will require a list not a string. What do you think?

I thik it's gonna lead to bugs to drop the support for string directly (i use it in every module that need a test dataset). But it's easy to test the type at the beggining and transform string into a list of string.

12rambau:
> * I used an automatic translation to french, so I've zero clues if that makes sense ;).

I'll correct the typos no worries

12rambau (approved):

PR #285, opened 2021-09-22
Reclassify branch

dfguerrerom:


12rambau (changes requested):
@dfguerrerom I see that codecov is complaining a lot and it was really a pain to test everything only at the end ;-) 

Now that we are using pytest (way much more pythonic than unittest) could you have a look and check if you can add the appropirate tests ? 

If you have questions and/or don't manage to find something I can help/answer

dfguerrerom:
- First, sorry for the first commit, of this round, it is just noise. 
- I've introduced some useful fixtures in the `conftest.py` file, could you take a look?, so I can keep testing the other objects?
- In the ReclassifyModel I got an 80% of coverage, the only method that I haven't tested is the reclassify with GEE... do we have to export the data and then wait for completion to test the result?

12rambau:
> * First, sorry for the first commit, of this round, it is just noise.

No worries, I make noise all the time 

> * I've introduced some useful fixtures in the `conftest.py` file, could you take a look?, so I can keep testing the other objects?

I discovered `pytest` recently as I was force by the Jupyter team to use it for the small feature I added in `ipyvue`. I found it more pythonic and turned everything in `sepal_ui` from `unittest` to `pytest`. But this `fixture` thing is awesome, I'll move some other fixtures that I'm creating over the other classes in here that will be way easier to read. 

In short yes, I like it, and yes please continue to use it !   

> * In the ReclassifyModel I got an 80% of coverage, the only method that I haven't tested is the reclassify with GEE... do we have to export the data and then wait for completion to test the result?

👏 👏 80% that's a super high score for a whole new module I just wanted you to test the new functions, not whole module !
for the reclassify with GEE yes you'll need to export it to check but this test will also need to be skipped for the github env ! (as we have no right to export with the authentification we use in actions). So if you don't test it I won't complain.

12rambau:
also I've corrected the doc issue in master so don't pay attention if the build fails

12rambau:
ok so I get whiy it's not working and that will be painful to solve. 
The main issue is that pytest.fixture ar eloaded no matter what for everyone and in our case a high number of them cannot be loaded in any env (obviously the elephant in the room is the gee init in Github that use Google service account). I tried to chain multiple decorator to skip them but they are loaded anyway. I think the best way to proceed is to identify the most used fixtures and put them in conftest.py.

For the rest stick to the class structure that make the fixture only available to the class using it (and there I can skip waht need to be skiped). 

As you've done extensive work on this one and as you don't have my account credential to test gee locally, I'll take over for the refactoring

12rambau (approved):

PR #286, opened 2021-09-23
feat: add a targets parameter to switch method

12rambau:


12rambau:
@dfguerrerom I think I'm too tired to understand what's wrong but maybe you'll find the issue: 

when running : 
```python 
from sepal_ui.scripts import utils as su 
import time

btn = sw.Btn('toto', 'mdi-folder')

@su.switch('disabled', 'loading', targets=[True, False])
def on_click(widget, event, data):
    time.sleep(5)
    return 

btn.on_event('click', on_click)

btn
```

I raise the following : 
> ---------------------------------------------------------------------------
> UnboundLocalError                         Traceback (most recent call last)
> ~/libs/sepal_ui/sepal_ui/scripts/utils.py in wrapper_switch(self, *args, **kwargs)
>    359 
>    360             if not len(targets):
>--> 361                 w = on_widgets[0] if len(on_widgets) else self
>    362                 targets = [getattr(w, p) for p in params]
>    363 
>
> UnboundLocalError: local variable 'targets' referenced before assignment

if you find a solution/idea be my guest

dfguerrerom:
That's happening because you're trying to replace a global variable (global to the inner function) in local space (the inner function), as soon as you try to assign the variable in the local space, the global one disappears in that scope and then the interpreter will say that the variable is referenced before the assignment. 
It happened to me when coding the loading decorator, as you can see here: https://github.com/12rambau/sepal_ui/blob/4521975034ebc60ff5ecd2c0188abadded4a261a/sepal_ui/scripts/utils.py#L244 

This is a minimal example to see what's happening there.

````Python
def func_wrap(param):
    
    def func():
        print('local func1:', dir()) 
    
    def func2():
        a = param
        print('local func2', dir())
        
    def func3():
        print('local func3', dir())
        if param:
            param = 1
    
    print('global:', dir())
    
    func()
    func2()
    func3()

dfguerrerom:
I've just changed the name of the variable in the last commit. I haven't see your workflow.

dfguerrerom (changes requested):
could we validate if someone's set a different number of targets and params? that's causing an index error, but that's not useful to see what's happening.

dfguerrerom (approved):

PR #288, opened 2021-09-29
feat: Copy to clipboard

### LINKED ISSUE #163: create a copy-to-clipboard widget
Everything is in the title !

12rambau:
# overview 

It finally works, the RPC way of calling javascripts methods directly from python will be super useful. We needed to wait for voila XX.12 to make javascript executable from dashboards (that was already working from Jupyter notebook and lab). 

## Note 

The javascript command used here (`document.execCommand("copy")`) is not working on every browser.... but as a first step I would like to make it available as it's working on the big 3: 
- chrome
- firefox
- opera

It will need further work to make it compatible with everything (hello IE7) but that's another story. I'll wait for people to complain.

Fix #163

### DIFF
--- a/docs/source/widgets/clip.rst
+++ b/docs/source/widgets/clip.rst
@@ -1,0 +1,43 @@
+Copy to clipboard
+=================
+
+Overview
+--------
+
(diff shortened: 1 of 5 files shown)


dfguerrerom (commented):
I don't have much to say here, but the Clip class name might not be entirely comprehensive for all (could even be confused with a gis geoprocessing)... what do you think if we change it to CopyToClip, or ClipCopy, or something similar?

12rambau:
> I don't have much to say here, but the Clip class name might not be entirely comprehensive for all (could even be confused with a gis geoprocessing)... what do you think if we change it to CopyToClip, or ClipCopy, or something similar?

I like `CopyToClip`

dfguerrerom (approved):

PR #293, opened 2021-10-01
Next string script

dfguerrerom:
Create a string followed by a consecutive underscore and number


12rambau (approved):
the build failed because of the GADM bug. now it will bug anyway as I've added new assets from the other PR, let's accept this one anyway

12rambau (commented):
if there are more than 9 files it will crash as you're indexing with -1. let's use split instead. 
I'll modify it in reclassify_branch

PR #331, opened 2021-11-06
refactor: use kwargs pop

12rambau:
- Avoid the duplication of parameters using an elegant and more pythonic method.
- force the values of the one that we don't want user to overwrite (`children` usually) 
- easyer to read 
- faster to execute as it doesn't require updating traits but just a dict

### DIFF
--- a/sepal_ui/sepalwidgets/alert.py
+++ b/sepal_ui/sepalwidgets/alert.py
@@ -18,7 +18,7 @@ class Divider(v.Divider, SepalWidget):
 
     Args:
         class\_ (str, optional): the initial color of the divider
-        kwargs (optional): any parameter from a v.Divider. if set  again, ['class_'] will be overwritten
+        kwargs (optional): any parameter from a v.Divider. if set, 'class_' will be overwritten.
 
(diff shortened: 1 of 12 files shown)


sepal_ui/sepalwidgets/alert.py
```diff
@@ -18,6 +18,7 @@ class Divider(v.Divider, SepalWidget):
 
     Args:
         class\_ (str, optional): the initial color of the divider
+        kwargs (optional): any parameter from a v.Divider. ['class_'] cannot be overwritten
```
dfguerrerom: why you're using brackets to the kwargs? this can be misunderstood

dfguerrerom (commented):
Just a question. Why you're using `pop` method instead of just overwriting the key with the default value?

sepal_ui/sepalwidgets/alert.py
12rambau: because for some there are multiple one, I'll change it to make it clearer

dfguerrerom (approved):

Issue #340, opened 2021-11-16
refactor members declaration in all classes


12rambau:
I discovered recently (https://stackoverflow.com/questions/9305751/how-to-force-ensure-class-attributes-are-a-specific-type) that you can est up class members out of the __init__ functions and give them a specific docstring and type. 

I'm working to complete the new documentation of the API content and I'll see if I can safely refactor member declaration to make it compliant with the latest PEP.

I'll also try to investigate to pros and the cons of using traitlets systematically

dfguerrerom:
What PEP are you referring to? Why we would want to force some specific types instead of doing the validation through traits?

12rambau:
PEP 242 If I trust the SO answers. But don't be fooled, there are no type enforcement here (as usual in Python), it's just to help linter and sphinx. ex: 

```python

class A()
   a: int
   b: str
```

dfguerrerom:
Anyways, if we would want to use the typing in the methods, we could check https://docs.python.org/3.8/library/typing.html

12rambau:
I remember hitting into this typing lib some tile ago, in the end I realized it's again (and as usual in Python) just indications. 
You have no idea how disappointed I was ;-)

PR #343, opened 2021-11-16
Improve lib documentation

12rambau:
The functions and classes were difficult to access from the doc, I was always ending directly in the source code to understand what I needed, that is the sign that something is not working. 

Adapting some other docs found on the web I came up with this handy `.. autosummary` directive which is extremely impressive. 

Unfortunately we cannot use the `auto-gen` CLI as our classes inherit from `ipyvuetify` templates. It's leading to ALL the `ipyvuetify` methods and attributes being melted in the doc. 

So I updated these files by hand to only document the attributes and method we need. Also I discovered that attribute support docstring if they are placed AFTER the declaration of the member (weird tbf). But to do so I needed to extract all the attributes from the __init__ method leading to this question: 
- https://github.com/12rambau/sepal_ui/issues/340

## DEMO

- [the previous alert page](https://sepal-ui.readthedocs.io/en/latest/modules/sepal_ui.sepalwidgets.html#module-sepal_ui.sepalwidgets.alert)
- [the new one](https://sepal-ui--343.org.readthedocs.build/en/343/modules/sepal_ui.sepalwidgets.Alert.html)

## NOTE

at the moment it's only a draft, I want to see how it looks like on readthedoc and I also need to debug some stuf I broke. I'll have a second round of refactoring on the member declaration to make it better.


dfguerrerom:
It looks really nice!

dfguerrerom (approved):

PR #346, opened 2021-11-18
make the viz member as a trait

### LINKED ISSUE #341: make viz member a trait
instead of calling `hide()`and `show()` in `SepalWidget`we could make viz a trait and simply set it to True or False to switch widget state. it will avoid hiding `if` conditional statement in apps. 

`hide` and `show` will of course remain accecible to avoid break in current apps

12rambau:
fix #341 

- set the viz member as a bool trait 
- call `hide`or `show`according to the value of `viz` trait 
- make the `toggle_viz` simply toggle the viz value (I think it should be deprecated at some point in the future

the only issue is that when you call `hide` (or `show`) and there is effectivelly a modification, it will be called twice:

1. byt the effective call to `hide`
2. by the callback when the viz value is changed

That is not costing any sensible time but if you have any idea on how to fix it ;-)

### DIFF
--- a/docs/source/widgets/sepal_widget.rst
+++ b/docs/source/widgets/sepal_widget.rst
@@ -4,7 +4,7 @@ SepalWidget
 Overview
 --------
 
-:code:`SepalWidget` is an abstract object that embed special classes, it can be used with any ipyvuetify widget component:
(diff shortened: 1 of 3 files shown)


dfguerrerom (approved):
No idea how can we avoid the double call to the observed method, I'll think about that.

PR #347, opened 2021-11-19
Map state

### LINKED ISSUE #345: display a loading statebar when a layer is loading in the map
## issue 
I experienced some issues with the FCDM module as the map is super slow to load (it uses a adaptative buffer, the more you zoom the slower it gets it's a nightmare). The consequence is that the user clicks frantically on the `zoom` `unzoom` btn and the layer never fully loads and they eventually break the app.

I've looked into the loading widgets that exists in ipyleaflet and that's not very elegant (from my perspective).

## current implementation

I came up with the idea of overwriting the `add_layer` method to add a `StateBar` at the top left corner for every layer and dynamically show them when the tile are reloading. overwritting the `add_layer` method is super versatile as it's the basic function called by all the others (`add_ee_layer`, `addLayer`, `add_raster_layer` ... etc). 

- user doesn't change anything in the way object are added to the map
- each time you add a layer, it adds a statebar writting "loading \<whatever\>"
- the statebare is shown whenever the layer is loading (thank you traitlets)

## questions

- should we remove them when the layer is removed (overwritting remove_layer as well) ?
- should the statebar always be displayed (and instead of changing viz, we could change loading trait) ?
- do you think it could be useful in general ? 

## demo 

Here is a gif demo and the the code I used. 

feature_demo

```python
from sepal_ui.mapping import SepalMap
from ipyleaflet import WidgetControl
import ee

ee.Initialize()

# create a custom statebar
class StateBar(sw.StateBar):
    
    def __init__(self, **kwargs):
        
        name = kwargs.pop("layer", "layer")
        kwargs["_metadata"] = {"layer": name} 
        
        super().__init__(**kwargs)
        
        self.msg = f"Loading {name}"
        self.loading = True
        
    def activate(self, change):
        
        if change['new']:
            self.show()
        else:
            self.hide()
            
        return

# define a custom Map class
class Map(SepalMap):
    
    layer_state_list = []
    
    def add_layer(self, l):
        
        # call the original function
        super().add_layer(l)
        
        # add a layer state object 
        state = StateBar(layer=l.name)
        self.layer_state_list += [state]
        self.add_control(WidgetControl(widget=state, position='topleft'))
        
        # link it to the layer state
        #layer = next(l for l in self.layers if l.name == name)
        l.observe(state.activate, "loading")
        
        return

# create the map and zoom on congo
test_map = Map()
test_map.zoom = 10
test_map.center = [5.703447982149503, 28.32275390625] 

# load hansen & al ddataset
dataset = ee.Image('UMD/hansen/global_forest_change_2020_v1_8')
treeCoverVisParam = {
  "bands": ['treecover2000'],
  "min": 0,
  "max": 100,
  "palette": ['black', 'green']
}
test_map.addLayer(dataset, treeCoverVisParam, 'tree cover')

# load the S2 RGB product (SR)
def maskS2clouds(image):

    qa = image.select('QA60');
    
    # Bits 10 and 11 are clouds and cirrus, respectively.
    cloudBitMask = 1 << 10;
    cirrusBitMask = 1 << 11;

    # Both flags should be set to zero, indicating clear conditions.
    mask = (
        qa
        .bitwiseAnd(cloudBitMask).eq(0)
        .And(qa.bitwiseAnd(cirrusBitMask).eq(0))
    )

    return image.updateMask(mask).divide(10000)

dataset = (
    ee.ImageCollection('COPERNICUS/S2_SR')
    .filterDate('2020-01-01', '2020-01-30')
    .filter(ee.Filter.lt('CLOUDY_PIXEL_PERCENTAGE',20))
    .map(maskS2clouds)
)

visualization = {
  "min": 0.0,
  "max": 0.3,
  "bands": ['B4', 'B3', 'B2'],
}

test_map.addLayer(dataset.mean(), visualization, 'RGB')

# display the map
test_map
```

12rambau:
Fix #345 

I've changed the behaviour of the statebar as dicussed, `state` is a permanent member of the map and will display the number of layers that are currently loading. 

I suggest you test it in your apps to check if does not lead to funny behaviour, As it implies multiple callbacks and we cannot set "prevent_default" I can only hope that we won't enter an infinite loop. 

I also make it invisible by default to not change the already existing maps.

I have issues with the tests, we never add layers on the test maps and I won't know how to test if the layer have changed state or not. I'll need to see how method mocking works around Pytest


dfguerrerom:
I just want to leave here what we discussed in our internal chat:
- I've tested on the SEPAFE app, when you load several images and you remove them quite fast, the spinning wheel remains loading forever and sometimes the number of loaded layers is higher than the actual value:

image

- In a previous PR some parameters were moved as class parameters to improve the automatic documentation of them, however, having mutable class arguments will lead to having shared properties in instances created from the same class, it was noticed when trying to create two maps with the `statebar` set as true, they will share this element, to overcome this situation we proposed:
  - Convert the new class parameters into traits, or
  - Give an init value for the class parameters within the initializing function

dfguerrerom (commented):
I just want to leave here what we discussed in our internal chat:
- I've tested on the SEPAFE app, when you load several images and you remove them quite fast, the spinning wheel remains loading forever and sometimes the number of loaded layers is higher than the actual value:

image

- In a previous PR some parameters were moved as class parameters to improve the automatic documentation of them, however, having mutable class arguments will lead to having shared properties in instances created from the same class, it was noticed when trying to create two maps with the `statebar` set as true, they will share this element, to overcome this situation we proposed:
  - Convert the new class parameters into traits, or
  - Give an init value for the class parameters within the initializing function

12rambau:
I think I now better understand how trigger works. 

Now the LayerStateControl is fully independant from the map and all the methods live there. 
I'm listening to the `layers` trait from the map and triggering 2 different behaviour when the number of layer change: 

1. if it's a new one I increase the number of layer by 1 and add an listener
1.  if it's a remove I'm decreasing the nb_layer and the nb_loading_layer if the layer was still loading when removed

It's more fluid from my side and avoid the previous issue when removing and adding layers too fast. @dfguerrerom let me know if you are still experiencing the same behaviour from your side

PR #348, opened 2021-11-19
build sepal_ui with ipyvuetify 1.7 and ipyvuetify 1.8.1

12rambau:
Build sepal_ui with new latest ipyvuetify version. It solve multiple problem such as: 
- markdown target to _blank
- overwrite all the vuetify objects
- use class management member


sepal_ui/sepalwidgets/sepalwidget.py
```diff
@@ -17,25 +17,57 @@ class SepalWidget(v.VuetifyWidget):
...
+
+            # change class value
+            self.class_list.remove("d-none")
+            self.old_class = str(self.class_)
+            self.class_ = "d-none"
 
         return
 
+    @deprecated(version="2.6.0", reason="Equivalent to :code:`self.viz = not self.viz`")
     def toggle_viz(self):
```
dfguerrerom, line 68: I think this method is simpler than changing the state by traits

dfguerrerom (commented):
So, to summarize, and if I understood this PR well...
In the next release, `hide` and `show` will be overwritten by the native ipyvue methods, but as those methods don't store the previous `class_`, we will end using the new `viz` trait and substitute the current implementation of the `.show` and `.hide` methods... right?

sepal_ui/sepalwidgets/sepalwidget.py
12rambau: then lets remove the deprecation notice

dfguerrerom:
Yes, I remember that specific use case, having `d-none` along with `d-flex` won't hide the widget... So just to clarify it in my mind, why would be better to use a trait to change the visualization state instead of the current working sepal_ui hide and show methods?

12rambau:
We had the viz member that was giving the information of the viz status, I thought that would also be nice to pilot it from there. 

The big advantage of `viz` being a trait is that you can now pilot visibility with things like: 

```python
w.viz = toto == tutu
```

instead of 
```python
if toto == tutu: 
   w.show()
else:
   w.hide()
```

note that I don't think we should drop `hide` and `show` as they are useful in many use cases and overwrite the native methods from ipyvue

dfguerrerom (approved):
Ok, so if we'll keep the methods, I think it's ok!

PR #384, opened 2022-01-21
use fontawesome instead of mdi icons

12rambau:
The font awesome libs is the one used by SEPAL nativelly. to make sure that the display is the same I decided to change what we use commonly in `sepal-ui` to match SEPAL main framework style.

- add an extra link in the html output (even though it's natively supported by voila, the display was funny in JupyterLab) 
- change every icon in the lib
- change the documentation accordingly

Could you test it from your side and let me know if anything behaves strangely?


dfguerrerom (approved):
I've tested the widgets and everything looks working properly ;)

PR #398, opened 2022-02-15
style: use relative paths

dfguerrerom:


12rambau (changes requested):
the readme is used in the documentation via an `include` directive (https://raw.githubusercontent.com/12rambau/sepal_ui/master/docs/source/start/readme.rst). If you use relative import then it will break the documentation page links as you can see in the pre-build of this PR (https://sepal-ui--398.org.readthedocs.build/en/398/start/readme.html).

Any specific reason why you want to change this behaviour ?

dfguerrerom:
(oops you were answering while I was writing)...
Well... actually this doesn't work...
The thing is that when you go to the [getting started page](https://sepal-ui.readthedocs.io/en/latest/start/getting_started.html), the `contributing guidelines` and `contributor` links at the bottom of the page, are pointing to the full raw `.rst` file in the github repo and is not using the `readthedocs` website link, I mean, I wanted to change it to avoid get out from the documentation website.
But instead if we use the relative path `docs/start/contribute.rst`, then the github readme will have the include page... hmm

12rambau:
Then maybe instead we could tweak the documentation structure a bit: 
Copy a readme as a landing page (without the badges as they prevent the rendering of the pdf version) 
use installation as the landing of the "getting started" section

dfguerrerom:
> 

Do you mean hard-copy the "read me" file instead of including it in the docs?

12rambau:
I’ve changed a bit the file structure: 
- removing `getting_started` to add an `index` page instead
- copy-paste the `README` in the landing page of the doc (removing the references to the documentation, the badges and the links to github) 

@dfguerrerom, I cannot request your review on your own PR, but if you're happy with these changes I've approved it so you can merge it whenever you feel like it.

PR #399, opened 2022-02-15
refactor: remove __setattr__ magic method.

dfguerrerom:
This method was added due to the well known and now fixed bug of v_slots in ipyvuetify. We do not need it anymore.


12rambau (changes requested):
Can you correct the failing test? I assume that as you're removing the function the related test is no longer needed.

you can run locally the problematic test by using: 
```
python -m pytest tests/test_Tooltip.py
```

dfguerrerom:
> Can you correct the failing test? I assume that as you're removing the function the related test is no longer needed.
> 
> you can run locally the problematic test by using:
> 
> ```
> python -m pytest tests/test_Tooltip.py
> ```

I forgot it. Done!

12rambau (approved):

12rambau:
who cares about .04% of coverage?

PR #413, opened 2022-03-04
manage translation with a btn in the app bar

12rambau:
## changes

- create a new config file for sepal-ui based app (`.sepal-ui-config`) written using .INI convention to store the default language of the user. This config file is applied to every app that supports it (sepal-ui >= 2.7.0)
- new `sw.LocaleSelect` object to set the language. It display language name in english + the flag + the language code using the same menu as the one used in `sw.FileSelect`.
- add automatically this widget in app bars (far right side) 
- when the language is changed, a alert is display on top of the app to tell the user to restart. 
- at app launch if the language requested is not available, the app will default to a related language (subvariety) or the default ("en")

## demo 

This is currently working with the template app (https://github.com/12rambau/sepal_ui_template) master branch : 

locale_teaser

## modification required in currents application 

- make sure that language folders are corresponding to official languages (https://github.com/12rambau/sepal_ui/blob/i18n/sepal_ui/scripts/locale.csv)
- set the app `cm` `sw.Translator`object to the `sw.AppBar` so that it initialize correctly 
- set the app `cm` `sw.Translator` object to the `sw.App` so that it display information when language is changed
- remove the loading of the env variable in the app `component.message/__init__.py` and then remove all parameter but the folder path
- you're ready to go 🚀 

The modification have been done in the master branch of the template: https://github.com/12rambau/sepal_ui_template/commit/20817e509d93ea6756e0160aa185df42d2ae534c


sepal_ui/sepalwidgets/app.py
```diff
@@ -380,3 +409,122 @@ def add_banner(self, msg, **kwargs):
...
+            close_on_content_click=True,
+            v_slots=[{"name": "activator", "variable": "x", "children": self.btn}],
+            value=loc.code,
+        )
+
+        # add js behaviour
+        jsdlink((self.language_list.children[0], "v_model"), (self, "value"))
+        self.language_list.children[0].observe(self._on_locale_select, "v_model")
+
+    def _get_country_items(self):
```
dfguerrerom, line 491: What do you think if we only get the available languages instead of listing all of them? it won't be useful for the users if they click and the language is not available.

12rambau: That was my initial thought but as it's a general parameter for all apps I decided it would be easier to make it always general. My idea is: it will only be set once in any app and then you rarely change the parameter again. If you have a different set of languages depending on the app, that can be disturbing e.g.:   
- your default language is "Albanian"
- App does not support it. You want to swith to another one but you are still not satisfied.
- then to switch back to "Albanian" you need to find in SEPAL the one and only app that support this language.

I agree though that the user should know which languages are supported in the app he's using. 
There are 3 informations to display: 
- what language is supported
- what is the current default
- what is the currently used one

maybe adding some chips in the list next to the flags ?

12rambau:
updates: 
- only display the available locales
- set the btn to "info" when before reload 
- merge any `.json` file in the l10n folder (allow to split the keys in multiple files) 
- create a small script to manually change value (that's for me and Andreas) when we do multiple presentation in 2 different language

dfguerrerom (approved):

PR #416, opened 2022-03-11
Drawer branch notification

### LINKED ISSUE #415: Interact with navigation drawers
Sometimes is useful to pass some data from the module model to the app environment and so far we do not have this implementation.

We can add two simple methods to the drawers so they can update their state with icons, badges, and so.

dfguerrerom:
- Aims to close #415
- The way to interact with the drawer is by passing a model and a trait, the drawer will be listening the model trait and will add/remove a red badge accordingly. 

Proyecto sin título

Let me know what do you think!

### DIFF
--- a/sepal_ui/sepalwidgets/app.py
+++ b/sepal_ui/sepalwidgets/app.py
@@ -84,6 +84,9 @@ class DrawerItem(v.ListItem, SepalWidget):
     alert = Bool(False).tag(sync=True)
     "Bool: trait to control visibility of an alert in the drawer item"
 
+    alert_badge = None
+    "v.ListItemAction: red circle to display in the drawer"
+
     def __init__(
         self,
         title,
@@ -123,6 +126,8 @@ class DrawerItem(v.ListItem, SepalWidget):
         # call the constructor
         super().__init__(**kwargs)
 
+        # cannot be set as a class member because it will be shared with all
+        # the other draweritems.
         self.alert_badge = v.ListItemAction(
             children=[v.Icon(children=["fas fa-circle"], x_small=True, color="red")]
         )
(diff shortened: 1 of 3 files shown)


sepal_ui/sepalwidgets/app.py
```diff
@@ -108,6 +126,46 @@ def __init__(self, title, icon=None, card=None, href=None, **kwargs):
         # call the constructor
         super().__init__(**kwargs)
 
+        # cannot be set as a class member because it will be shared with all
+        # the other draweritems.
+        self.alert_badge = v.ListItemAction(
+            children=[v.Icon(children=["fas fa-circle"], x_small=True, color="red")]
+        )
+
+        if model:
```
12rambau, line 135: Could it be squeezed to `if model and not bind_var` ?

sepal_ui/sepalwidgets/app.py
```diff
@@ -108,6 +126,46 @@ def __init__(self, title, icon=None, card=None, href=None, **kwargs):
...
+            if not bind_var:
+                raise Exception(
+                    "You have selected a model, you need a trait to bind with drawer."
+                )
+
+            link((model, bind_var), (self, "alert"))
+
+        self.observe(self.add_notif, "alert")
```
12rambau, line 143: instead of using the function, could you do the same with a decorator?

sepal_ui/sepalwidgets/app.py
dfguerrerom: The `if` statement is not only used to trigger the exception, if the model is set, we link the `traits`, so we don't need to use the `else`.

12rambau (approved):

PR #419, opened 2022-03-15
create a bnt to switch between light and dark theme

### LINKED ISSUE #139: automatically detect the theme used and adapt the frontend accordingly
I have seen it on several  zoom : it's so cool to display everything in dark but look so wrong when people are actually just developing in there own env in light theme. 

Several situation that I would like to handle : 
- voila theme
- jupyterLab theme 
- jupyter notebook theme

12rambau:
Fix #139 

## changes

Inspired by the language btn I created the same interface to switch permanently between light and dark theme. 

- new light-dark switcher in the appBar
- display information message on chage 
- manage theme via the config_parser
- change default basemap of the map with respect to the selected theme 
- create 3 extra colors to manage bars and menu (a dark grey background to maintain the depth even in dark theme)
- change files icons colors for light theme 
- create a script (because I like script) to do the same as the btn

## modification required in current application 

simply make sure that colors are coming from `sepal_ui.color` instead of `v.theme.themes.primary`


sepal_ui/sepalwidgets/app.py
```diff
@@ -461,12 +482,25 @@ def add_banner(self, msg, **kwargs):
...
         kwargs["dismissible"] = kwargs.pop("dismissible", True)
         kwargs["children"] = [msg]  # cannot be overwritten
+        kwargs["_metadata"] = {"id_": id_}
 
-        # create the alert
         alert = v.Alert(**kwargs)
 
-        # add the alert to the app
-        self.content.children = [alert] + self.content.children.copy()
+        # Verify if alert is already in the app.
```
dfguerrerom, line 489: If you click several times the `theme` or `locale` button, you'll end with a lot of banners with the same info.
I've added this verification to avoid multiple aggregations, what do you think? perhaps there is a cleaner way to avoid those nested if's-

sepal_ui/sepalwidgets/app.py
```diff
@@ -600,3 +643,62 @@ def _on_locale_select(self, change):
...
+        """
+        toggle the btn icon from dark to light and adapt the configuration file at the same time
+        """
+        # use a cycle to go through the themes
+        theme_cycle = cycle(self.THEME_ICONS.keys())
+        next(t for t in theme_cycle if t == self.theme)
+        self.theme = next(t for t in theme_cycle)
+
+        # change icon
+        self.color = "info"
```
dfguerrerom, line 695: Can we also change the button color? depending on the current theme?

sepal_ui/sepalwidgets/app.py
12rambau: Agreed, note that if we do that then "locale" and "theme" need to become protected `id_` for Tiles (as they are also children of app.) 

I'm trying something with `next` and a single `try ... except` let's see how it goes

dfguerrerom: What do you mean with "become protected `id_`"?, Tile cards have different metadata key in the dictionary so I don't see how they can be affected.

sepal_ui/sepalwidgets/app.py
12rambau: sure what do you have in mind (colors are proposition I'm pretty sure there will be modifications) ?

sepal_ui/sepalwidgets/app.py
12rambau: true, I'm positive to COVID I'm less sharp than usual but I would like to finish this PR before going to bed to hibernate 😷

12rambau: any better since last push ?

sepal_ui/sepalwidgets/app.py
```diff
@@ -457,16 +478,31 @@ def add_banner(self, msg, **kwargs):
...
+        kwargs["transition"] = kwargs.pop("transition", "scroll-x-transition")
         kwargs["prominent"] = kwargs.pop("prominent", True)
         kwargs["dismissible"] = kwargs.pop("dismissible", True)
         kwargs["children"] = [msg]  # cannot be overwritten
+        kwargs["_metadata"] = {"id_": id_}
+        kwargs["v_model"] = kwargs.pop("v_model", False)
+
+        # prepare a test to find an existing banner
+        def test(m):
+            m is not None and m.pop("id_", None) == id_
```
dfguerrerom, line 490: You're missing a `return` here.
And if you click `theme`, then `locale`, and `theme` button again, new widgets will be displayed.

12rambau: that was unexpected but I found the feature nice in the end. Like as long as you click on the same btn we continue to refresh the alert, if you click elsewhere, we "forget" it and reset the alert again. that's my 2 cents, if I'm alone I agree to change it.

dfguerrerom: Well... the nested if-solution was working pretty well tho, I'll let you chose the better one ;).

sepal_ui/sepalwidgets/app.py
dfguerrerom: Well, let's leave this conversation as it is, and later we can check when agreeing with a new light theme proposition.

sepal_ui/sepalwidgets/app.py
12rambau: too muche resonsability 😄

12rambau: I think I found it: 
- attributes is always a dictionary even when not set so I can directly use the `get` method without checking existence
- if `alert` is set then I remove it else I do nothing (the use of the `or` is more elegant and don't create indentation
- prevent the previous bug

dfguerrerom: Perfect!! It worked pretty well, I won't bother you anymore with this!

PR #430, opened 2022-03-25
Command script to activate virtual envs

dfguerrerom:
This command script is intended to make easier the search and activation process of the virtual envs in the platform.

Basically, it will look at the installed venvs in both usr and local module-venvs, and you can choose which one you want to activate.

what do you think? I found a little bit different way to activate the ven but it is working, do you mind taking a review?

### DIFF
--- a/sepal_ui/bin/activate_venv
+++ b/sepal_ui/bin/activate_venv
@@ -22,7 +22,6 @@ if __name__ == "__main__":
             for line in result.stdout.decode("utf-8").splitlines()
             if ".local" not in line
         ],
-        # columns=["name", "path"]
     ).iloc[1:]
 
     # Check if user has test envs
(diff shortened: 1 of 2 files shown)


sepal_ui/bin/activate_venv
```diff
@@ -0,0 +1,80 @@
...
+            if ".local" not in line
+        ],
+    ).iloc[1:]
+
+    # Check if user has test envs
+    test_venv_path = Path.home() / "module-venv"
+    if test_venv_path.exists():
+        test_envs = pd.DataFrame(
+            list(
+                [f"test {el.stem}", str(el)]
```
12rambau, line 32: I think you should use `el.name` to catch venv name that include a "dot" like "se.plan"

12rambau (approved):

PR #439, opened 2022-04-05
feat: create planet handler

dfguerrerom:
I think that we could merge the different planet calls into one class, so it can be shared among planet depending apps, what do you think?

I'm opening a request with a simple object that we can expand depending on our needs, for example, query base maps, daily imagery, and so on, please let me know what do you think so we can keep building.

I haven't set the test since this is a wip, and is open to ideas.


dfguerrerom:
Ooops not passing due to the GADM server issue

12rambau (approved):
nothing to add, we will surely add some stuff afterward.

Excellent idea.
We could test it using my key ?

dfguerrerom:
lasts commits include:
- Planet view to handle credentials (user+pwd) or API key inputs.
- Errors are handled directly from the model, not the view, so the model can be used without a view.
- Tests to planet model functions
- Alert and button are optional elements

Once you review the changes, I'll work on:

- Documentation
- View tests


Please @12rambau  let me know what do you think and what can be changed/improved.

dfguerrerom:
The tests are working on sepal because there is a legacy `.planet.json` file in usr, however, when this file is not present, an error is being triggered when the `client` is initialized (the build).
I have seen that there is a new authentication workflow in the planet repo https://github.com/planetlabs/planet-client-python/blob/main/planet/auth.py, so I think the best option would be to wait until there is a new release...

dfguerrerom:
> > The tests are working on sepal because there is a legacy `.planet.json` file in usr, however, when this file is not present, an error is being triggered when the `client` is initialized (the build). I have seen that there is a new authentication workflow in the planet repo https://github.com/planetlabs/planet-client-python/blob/main/planet/auth.py, so I think the best option would be to wait until there is a new release...
> 
> But if it's only the credential authentication that fail, you can also comment it and only provide support for the API key for the time being? And whenever the option becomes available you uncomment both code and test


The credential (user+pwd) authentication is not the reason for the fail... the reason is that the current planet authentication workflow is not throwing an error when the client is instantiated without a "key", making it difficult to realize why Sepal was working (https://github.com/12rambau/sepal_ui/pull/439#issuecomment-1121538658) and here it doesn't. 

I've based the model on the current planet version, taking advantage of all its functionality (even though I realized that the auth is not straightforward)... with the last commit (5070f876a01f10c70b00f2934fe7a6156486be92), the tests will pass and this component will be ready to use.

I wanted to stop the development because I realized that the planet repository has been working in the last week within the authentication process, making it totally incompatible with this version, however, it's true that we can use this version in the time being and adapt it later when they release latest changes, so thanks for the suggestion.

PD: the last commit is also failing, but my best guess is that the secrets API key is not valid.

sepal_ui/planetapi/planet_view.py
```diff
@@ -0,0 +1,120 @@
+import sepal_ui.scripts.utils as su
+from traitlets import observe, dlink, Bool
+from planet_model import PlanetModel
+import ipyvuetify as v
+import sepal_ui.sepalwidgets as sw
+
+
+class StateIcon(sw.Tooltip):
```
12rambau, line 8: Don't you think this very one could be useful as a general component in sw ?

could you also add a super simple test where you just init the component (so that codecov stops complaining)

12rambau (commented):
I really like it! API documentation still needs some adjustments and I wrote few comments

sepal_ui/planetapi/planet_view.py
dfguerrerom: I've tested on my side in a fresh env and it worked as it was working before, just `command`, so I think that entry point is managing this... 
Now that they are python scripts, I think we can make specific tests, and validate them by capturing their `stdout`, what do you think?

12rambau: I think you answered in the wrong PR ;-)

sepal_ui/planetapi/planet_view.py
```diff
@@ -0,0 +1,111 @@
...
+        self.w_key.v_model = None
+
+    def _swap_inputs(self, change):
+        """Swap between credentials and api key inputs"""
+
+        self.planet_model.init_client(None)
+        self.alert.reset()
+        self.reset()
+
+        if change["new"] == "api_key":
```
12rambau, line 93: from what I understand if it's a swap it means that I cannot select the same value twice. what about instead of a if statement we use the `toggle_viz()` method from SepalWidget to change the 3 components status ?

sepal_ui/planetapi/planet_model.py
```diff
@@ -0,0 +1,136 @@
...
+        if the method is called from an event (view)
+
+        Args:
+            credentials (str, tuple): planet API key or tuple of username and password of planet explorer.
+            event (bool): whether to initialize from an event or not.
+        """
+
+        credentials_ = credentials
+        if event and not any(tuple(credentials)):
+            raise ValueError("Please fill the required field(s).")
```
12rambau, line 48: all these error messages will be send directly to the end-user. Could you add specific keys to the `messages` so that they can be translated?

12rambau (approved):
you inverted the order of the items so that credential becomes the default entry point  right ? 

for me it's ++ and good to go! 🎊

dfguerrerom:
> you inverted the order of the items so that credential becomes the default entry point right ?
> 
> for me it's ++ and good to go! 🎊

Actually it was already the default, but in the dropdown list it was the second.

Issue #425, opened 2022-03-23
Create a default feature visualization on map


dfguerrerom:
When I want to rapidly visualize a feature layer in the `SepalMap` I would like to do `map_.addLayer(layer)` and visualize it directly into the map, however, as the default visualization in `geemap` is black and our default base map is black, you can't see the layer.

Instead of always setting a default visualization to the feature collections in the map for every app, it would be nice to have at least a default visualization for each of the default themes, of course as an optional parameter that can be overridden.

The current alternatives could be to keep using the own visualization in each module, but at the end to keep the consistency we could just add this behavior to `add_layer` and `addLayer`.

What do you think?

dfguerrerom:
I'm reopening it because the `ee.FeatureCollection` layers are not displaying the default style, it's only working with images or local tiles.

```python
from sepal_ui.mapping import SepalMap
import ee
ee.Initialize()

geometry = ee.FeatureCollection(ee.Geometry.Polygon(
        [[[-103.198046875, 36.866172202843465],
          [-103.198046875, 34.655531078083534],
          [-100.385546875, 34.655531078083534],
          [-100.385546875, 36.866172202843465]]],
))
map_ = SepalMap()
map_.addLayer(geometry)
map_.centerObject(geometry)
map_

```

PR #441, opened 2022-04-05
Create a default style to SepalMap add_layer method

dfguerrerom:
Aims to close https://github.com/12rambau/sepal_ui/issues/425.

Basically, I've done two things:

- Move style calls from `__init__` to `frontend.styles` because colors were not accessible from an upper level, and imported them there to avoid breaks!
- Override SepalMap `add_layer` method to use a default style to the GeoJSON objects when there is not one specified.


sepal_ui/frontend/styles.py
```diff
@@ -93,3 +141,24 @@ class Styles(v.VuetifyTemplate):
     "DEFAULT": _other,
     "PARENT": _parent,
 }
+
+
+# Set default styles to layers
+
+layer_style = {  # default styling of the layer
```
12rambau, line 148: could you change the comments (I think they are coming from the alert module and they make no sens in this file)

sepal_ui/mapping/mapping.py
```diff
@@ -782,3 +784,20 @@ def get_viz_params(image):
...
+        """Add layer and use a default style for the GeoJSON inputs
+
+        default_hover (bool): whether to use the default hover style or not.
+
+        """
+
+        if isinstance(layer, GeoJSON):
+
+            if not layer.style:
+                layer.style = styles.layer_style
```
12rambau, line 798: for coverage purpose, could you add a single test, with a fake geoJson (like a triangle anywhere) so that these 2 if statement are used at least once in the test suit

sepal_ui/__init__.py
```diff
@@ -1,56 +1,11 @@
...
-from configparser import ConfigParser
+from sepal_ui.frontend.styles import color, get_theme, theme
 
 import ipyvuetify as v
 
-from sepal_ui.frontend import styles
-
 __author__ = """Pierrick Rambaud"""
 __email__ = "pierrick.rambaud49@gmail.com"
 __version__ = "2.7.0"
```
12rambau, line 9: you confirm that it's still possible to call color as such: 

```
import sepal_ui 

sepal_ui.color.primary
>>> the selected color hex code
```

I'm in an airport it's complex to run sepal

12rambau (changes requested):

sepal_ui/__init__.py
dfguerrerom: Yes, that will work since we are importing `color` and the other required functions (I've tested it).

sepal_ui/mapping/mapping.py
```diff
@@ -782,3 +784,22 @@ def get_viz_params(image):
                     props = {}
 
         return props
+
+    def add_layer(self, layer, default_hover=False):
```
12rambau, line 788: what about changing `default_hover` to a simpler `hover` parameter ?

dfguerrerom: I've actually changed from `hover` to `default_hover` in my last commit... I agree that it can be useless.

12rambau (approved):

Issue #443, opened 2022-04-14
get and set widget children


dfguerrerom:
While developing some widgets we have realized that one recurrent task is to look for a specific children's widget and set a new one based on indexes, these methods can be implemented on SepalWidget.

PR #447, opened 2022-04-15
Sepalwidget

dfguerrerom:
In relation to https://github.com/12rambau/sepal_ui/issues/443.

- Rename SepalWidget overwrote Vuetify Classes
- Two new methods created `set_children` and `get_children`:
  - I've only included `first` and `last` as available positions when setting new children, do you think is enough? or should we use the indexes instead?
- Tests
- Docs


12rambau (approved):

12rambau:
I really like it. I don't have any specific use case in mind so let's merge it and if we really need to use index value we will do another PR

PR #461, opened 2022-05-04
docs: add CLI tools

### LINKED ISSUE #444: add CLI to the documentation
everything is in the title

12rambau:
Fix #444

### DIFF
--- a/docs/source/cli/index.rst
+++ b/docs/source/cli/index.rst
@@ -1,0 +1,24 @@
+CLI
+===
+
+    command line tool 
+    
+:code:`sepal-ui` provides several command line utilities that will help the developer to manipulate its applications and environment.
+
+.. autoprogram:: module_deploy:p
+   :prog: module_deploy
+   
+.. autoprogram:: module_factory:p
+   :prog: module_factory
+   
+.. autoprogram:: module_l10n:p
+   :prog: module_l10n
+   
+.. autoprogram:: module_theme:p
+   :prog: module_theme
+   
+.. autoprogram:: module_venv:p
+   :prog: module_venv
+   
+.. autoprogram:: activate_venv:p
+   :prog: activate_venv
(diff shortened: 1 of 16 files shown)


dfguerrerom:
I'm helping out with this doc build which was failing. 
Basically, the last one was failing because `bin` folder was not in the `sys.path`, and therefore the cli modules were inaccessible for the docs build, however, it seems like `autoprogram` doesn't do what we expected to do when the cli are installed as scripts with shebang, so I've changed them to python scripts —I've noted that is most frequent in that way. 
Now, I suppose that we have to add missing packages that are imported in the cli modules to make the build work, what do you think?

12rambau:
- add the bin folder to sys path 🎊 
- replace `p` by `parser`, that's a naming convention but you're right it's easier to get what it's about.

On the python scripts I'm curious, will it be still possible to do:
```
$ module_factory
```

or will we need to run 
```
$ python -m module_factory
```

?

dfguerrerom:
> * add the bin folder to sys path 🎊
> * replace `p` by `parser`, that's a naming convention but you're right it's easier to get what it's about.
> 
> On the python scripts I'm curious, will it be still possible to do:
> 
> ```
> $ module_factory
> ```
> 
> or will we need to run
> 
> ```
> $ python -m module_factory
> ```
> 
> ?

I've tested on my side in a fresh env and it worked as it was working before, just $ command, so I think that entry point is managing this...
Now that they are python scripts, I think we can make specific tests, and validate them by capturing their stdout, what do you think?

(moved reply to this thread)

12rambau:
> Now that they are python scripts, I think we can make specific tests, and validate them by capturing their stdout, what do you think?

I would say test will be for another PR, but agreed

dfguerrerom (approved):
Do you think we can merge it now?

PR #472, opened 2022-05-14
fix: closes #466

### LINKED ISSUE #466: name is not set when selecting a default admin in the AoiTile
using the following code: 

```python
aoi_tile = aoi.AoiTile(admin=100)
aoi_tile.view.model.gdf is None
>>> False
```

I generate an aoiTile already set on Guadeloupe. Problem, the name is not set automatically:

```python
aoi_tile.view.model.name
>>> "!!disabled!!"
```

I think there's some conflict between 2 events of the Tile/View

dfguerrerom:


12rambau (approved):
excellent thanks!

let me just take care of the patching errors and I'll merge it

PR #482, opened 2022-05-30
Tooltip

### LINKED ISSUE #342: add a `add_tooltip` method
I don't know where to start from, tooltip syntax is super tedious and I would like to find a way to make it easier. 

My first thought is: it's mainly useful for btn so I was thinking of adding a add_tooltip method to the btn object. that would return the tooltip (and not the btn) and that's where it can be tricky. 

Also it could be added to `SepalWidget` but I' would argue that nobody will add a tooltip to a `sw.Tile` object.

I'm open to discussion, this issue is just a reminder

12rambau:
This aims to close #342.

## overview

- I moved `Tooltip` in the `sepalwidget` file to avoid circular referencing not changing anything from the outside as everything is called in the `__init__`.
- I created a `set_tooltip` method also called in the constructor. The idea is to update the `Tooltip` text if it already exists or create one if necessary. 

## advantage 

The tooltip is now fully integrated in the widget. To avoid confusion for newcomers there are 2 way sof using a sepalwidget: 
```
# if not tooltip are set 
btn = sw.Btn()
btn 
```

```
# if you decided to display a tooltip
btn = sw.Btn("toto")
btn.widget
```

I don't know why btn is not visible surely a bug. I decided not to call the member "tooltip" to avoid confusion. Once a tooltip is added you can't remove it (or at least I didn't manged) so you'll be force to display it with `btn.widget` forever but you can still modify the message and even make it disapear by using `set_tooltip()`. 

While a widget is shown with a tooltip all the other interaction with the app (click event, jslink etc) can still be managed by the initial widget (`btn` in our example) avoiding the confusion of calling the widget FROM the tooltip.

Let me know if you have question, bugs, or suggestion on this one, I'm just trying to make it as easy as possible to use tooltip the last application realy need some.


dfguerrerom:
The widget alone is disappearing because it's activated by the `v_slot` of the tooltip... anyways I don't think is something harmful... if you create a tooltip, you for sure want to display the `widget.with_tooltip` in the view...

dfguerrerom (approved):
Please check my last changes and if you agree with them, we can merge in master.

12rambau:
> Please check my last changes and if you agree with them, we can merge in master.

I like the modification you are good to go

PR #490, opened 2022-06-01
feat: extend color simplenamspace to interactively display colors in …

dfguerrerom:
I always find myself struggling and remembering what are the sepal_ui colors. With this simplename extension, it will be easy to catch them.

image

### Update:

I've been playing with the styles structure and now there are some proposed changes:
- The `NameSpace` `Color` object now has traits. It is really useful when you want to change the current theme with just one line: `color._dark_theme = no color._dark_theme`.
- Even though we can change the theme by triggering this event, the new color references (i.e. `bg` , `menu`, ...) can't change automatically due to are, not traits (see image below), so the app has to be restarted anyways. And this point would be something to the future investigation, I think it would be possible.
- All remain working as it use to work before.

colors


sepal_ui/frontend/styles.py
```diff
@@ -1,112 +1,121 @@
...
 
-        super().__init__(**kwargs)
+        # Even if the theme.themes.dark_theme trait could trigger the change on all elms
+        # we have to replace the default values everytime:
+        theme = getattr(v.theme.themes, self.theme_name)
 
-    def as_hex(self):
-        """Return a color palette with hex codes instead of RGB values."""
-        return [mpl.colors.rgb2hex(color) for name, color in self.__dict__.items()]
+        # TODO: Would be awesome to find a way to create traits for the new colors and
```
12rambau, line 82: aggreed and it will also be super useful to make Model more userfriendly

PR #491, opened 2022-06-01
Create a AoiControl to zoom on different sites on the map

### LINKED ISSUE #389: create a zoom_to_aoi btn on the map
It could be a simple button that once set zoom to specific coordinates. (I have a terrible mouse and I get lost quite often)

From the dev point of view it's simply a value that need to be updated in the map when a layer is added.

I don't know if it could b useful from sepal-ui or if it's just something I should implement where I need it.

12rambau:
## Overview 

Fix #389 

Create a AoiControl as describe in the issue. 
When a dev create a new layer it can choose to add it to the `AoiControl` so that the user can navigate through it. particularly useful in application using Sub-AOIs.

## test code
```python
from sepal_ui import mapping as sm 
import ee
from shapely import geometry as sg 
from ipyleaflet import GeoJSON

ee.Initialize()

map = sm.SepalMap()
control = sm.AoiControl(map)
map.add_control(control)

point = sg.Point(0,0)
geo_json = GeoJSON(data=point.__geo_interface__, name="point")
map.add_layer(geo_json)
control.add_aoi("point", point)

ee_point = ee.Geometry.Point(0, 10)
map.addLayer(ee_point, {"color": "red"}, "ee_point")
control.add_aoi("ee_point", ee_point)

map
```


dfguerrerom:
Not sure if it is harming something but when I click over the layer two times in row, the following error arose:

```Python

File ~/sepal_ui/sepal_ui/mapping/sepal_map.py:206, in SepalMap.zoom_bounds(self, bounds, zoom_out)
    194 def zoom_bounds(self, bounds, zoom_out=1):
    195     """
    196     Adapt the zoom to the given bounds. and center the image.
    197 
   (...)
    203         self
    204     """
--> 206     minx, miny, maxx, maxy = bounds
    208     # Center map to the centroid of the layer(s)
    209     self.center = [(maxy - miny) / 2 + miny, (maxx - minx) / 2 + minx]

TypeError: cannot unpack non-iterable NoneType object



```

12rambau:
> Not sure if it is harming something but when I click over the layer two times in row, the following error arose:

When you click twice on the same element, you are in fact unselecting it => `v_model = None`.

I changed the control to move it on a "click" event. so now you can click move away manually and click again on the same one.

sepal_ui/mapping/aoi_control.py
```diff
@@ -0,0 +1,167 @@
...
+        self.m = m
+
+        # init the aoi data list
+        self.aoi_bounds = {}
+
+        # set some default parameters
+        kwargs["position"] = kwargs.pop("position", "topright")
+
+        # create a hoverable btn
+        btn = MapBtn(logo="fas fa-at", v_on="menu.on")
```
dfguerrerom, line 39: What about `search-location` or `search-plus`? `@` (as always) is not so intuitive for me.

dfguerrerom (commented):
I like this one, it's working perfectly on my side... I have some comments...

- Can we add "as_aoi" (or "with_zoom")  parameter to the "add_layer/addLayer" method? if the aoi_control is not on the map, we can create it, and if it is, just add layer_name + bounds?
- Also, if an aoi_layer (or "with_zoom" layer) is removed from the map, can we remove it from the control (without an extra call).

Perhaps we can have a Dict trait in the map ({name:[bounds]}), overwritten every time a new "as_aoi" layer is added, and an Aoi Control, listening every time to add/remove layers "as_aoi"

what do you think??

sepal_ui/mapping/aoi_control.py
12rambau: `search-location` makes complete sense and is aligned with what they use in the main framework

12rambau:
> Can we add "as_aoi" (or "with_zoom") parameter to the "add_layer/addLayer" method?

I tried to wrap my head around an easy way to deal with the "as_aoi" option but I think it's an overkill at the moment. 
There are 25 different layer type provided by ipyleaflet (https://ipyleaflet.readthedocs.io/en/latest/layers/index.html) and each one as a different way of storing its footprint (when it's not a point that need buffering). Implementing this functionality will have 3 consequences:
1. make the `add_layer` heavier as it will need to check for `zoom_control` every time
1. make the `add_layer` clumsy as if you set a non-suported layer type with option `as_aoi` I'll be forced to raise an error 
1. make the `SepalMap` dependent on its widget. I'd like to avoid going in the `geemap` direction and let the user free of choosing which control they want to add

dfguerrerom:
> > Can we add "as_aoi" (or "with_zoom") parameter to the "add_layer/addLayer" method?
> 
> 
> 1. make the `add_layer` heavier as it will need to check for `zoom_control` every time
> 2. make the `add_layer` clumsy as if you set a non-suported layer type with option `as_aoi` I'll be forced to raise an error
> 3. make the `SepalMap` dependent on its widget. I'd like to avoid going in the `geemap` direction and let the user free of choosing which control they want to add

Let's then keep this as it is but... I understand the problem when dealing with more layer types, even though we are mostly using 3-4 of them, so 

1. heavier yes, if it's compared with nothing (∞), but I wouldn't say that is something perceptible in the end... 
2. yes, however, we are mainly using some of these layers and would be the same if you do `control.add_aoi("heatmap", Heatmap)`.
3. there are pros and cons, but fewer lines of code, and less documentation to read if you just `add_layer(as_aoi=True)`.

dfguerrerom (approved):

Issue #294, opened 2021-10-05
alert update progress affects the time performance of an execution 


dfguerrerom:
The `update_progress` method of the `Alert` widget is impacting the performance of an app due to it has to render all the view every time it receives progress...

Consider the following example:

```Python

import sepal_ui.sepalwidgets as sw

alert = sw.Alert().show()
alert

%%time
progress = 0
def update_progress(b, bsize, tsize):
    alert.update_progress(min(int(b*bsize*100/tsize), 100)/100)
urllib.request.urlretrieve(url, Path('~').expanduser()/'tmp/dum.zip', update_progress)

```

The whole execution time of this snippet is around 20 seconds, while the following example only takes around 2.5 secs, that's 8x times faster.

```Python

import urllib
from tqdm.auto import tqdm
from pathlib import Path

url = 'https://firms.modaps.eosdis.nasa.gov/data/country/zips/modis_2000_all_countries.zip'

%%time
with tqdm(
    unit="B",
    unit_scale=True,
    unit_divisor=1024,
    miniters=1,
    desc="downloading...",
    bar_format="{l_bar}{bar}{n_fmt}/{total_fmt}",
    dynamic_ncols=True,
    total=urllib.request.urlopen(url).length,
) as f:

    def callback(b, bsize, tsize):
        f.update(b * bsize - f.n)

    urllib.request.urlretrieve(url, Path('~').expanduser()/'tmp/dum.zip', callback)

    f.total = f.n
```

This is a dummy process that takes around 500 steps to be fully completed, but, others could have more than that and its performance could be highly impacted... what do you think if we incorporate the `tqdm` bar in the `update_progress` method? could be easily included.

PR #495, opened 2022-06-03
feat: introduce tqdm progress bar. related with #294

dfguerrerom:
This PR aims to introduce the tqdm alert progress bar referenced by the (quite old)  #294 issue. 

- Even though it's doing what's expected, I can't manage to let the `tqdm` container `HBox` inherit the ipyvuetify styles, causing the text within the progress bar to be totally different. @12rambau would you like to give a check?
- In relation to the time, it does totally faster than the previous implementation.
- I've marked it as draft until we can manage the styles, and then I'll proceed with documentation and styles.

## Testing code:

```Python
import sepal_ui.sepalwidgets as sw
import urllib
from pathlib import Path
alert = sw.Alert().show()
url = 'https://firms.modaps.eosdis.nasa.gov/data/country/zips/modis_2000_all_countries.zip'
display(alert)

progress = 0
def update_progress(b, bsize, tsize):
    alert.update_progress(min(int(b*bsize*100/tsize), 100)/100)
urllib.request.urlretrieve(url, Path('~').expanduser()/'tmp/dum.zip', update_progress)

````

## Preview

Proyecto sin título

### DIFF
--- a/sepal_ui/frontend/styles.py
+++ b/sepal_ui/frontend/styles.py
@@ -128,6 +128,7 @@ class Styles(v.VuetifyTemplate):
     - remove shadow of widget-control
     - remove padding of the main content
     - load fontawsome as a resource
+    - ensure that tqdm bars are using a transparent background when displayed in an alert
     """
 
     template = Unicode(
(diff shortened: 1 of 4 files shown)


12rambau:
1. I change the color of the messages in the tqdm Box. It's inheritting the one from the alert itself. 
1. I also removed the background it was looking strange in some colors in light theme
1. I stopped displaying 100 in green (that's also weird on some colors)

if you think .2 is not looking nice feel free to reset to normal styling. for .3 maybe we could add a parameter to tell if you want to display the end in green ?

dfguerrerom:
Thanks, @12rambau. It's looking great. Previously made apps shouldn't have any incompatibility with this update when use the latest version.

12rambau (approved):

PR #498, opened 2022-06-06
Manage edited geometries

### LINKED ISSUE #386: customize the drawControl object to manage edition (not natively set up)
The ipyleaflet drawingControl cannot handle edit images not that it can produce reasonable geojson geometries linked to circles. 

Some solution have bee proposed here: https://github.com/jupyter-widgets/ipyleaflet/issues/849#issuecomment-1025069614

It would make sense to have a look

12rambau:
Fix #386 

- add a to_json method to the Draw_control. using it you'll get everything drawn on the map without the style as a geo_interface
- also use the previously coded (aoiView) polygonize method to manage circles 
- update to ipyleaflet 0.14.0 (introduction of the DrawControl data trait) 
- small correction of the remov_layer method (avoid throwing an error when none_ok == True)
- update AoiView to use this DC methods 
- refactor AoiView

```python 
from sepal_ui.mapping import *

m = SepalMap(dc=True)
m
```

Edit the features as much a you want and observe the content from 

```python 
m.dc.data
```

Export it as a usable geo_interface: 
```python
m.dc.to_json()
```

### DIFF
--- a/sepal_ui/mapping/draw_control.py
+++ b/sepal_ui/mapping/draw_control.py
@@ -1,4 +1,6 @@
+from shapely import geometry as sg
 from ipyleaflet import DrawControl
+import geopandas as gpd
 
 from sepal_ui import color
 
@@ -19,7 +21,6 @@ class DrawControl(DrawControl):
 
         # set some default parameters
         options = {"shapeOptions": {"color": color.info}}
-        kwargs["edit"] = kwargs.pop("edit", False)
         kwargs["marker"] = kwargs.pop("marker", {})
         kwargs["circlemarker"] = kwargs.pop("circlemarker", {})
         kwargs["polyline"] = kwargs.pop("polyline", {})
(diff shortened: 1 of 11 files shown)


dfguerrerom (changes requested):
- Now the "asset" name is displayed by default, can we hide it at the tile initialization? (as was before).
- I'm not able now to select administrative layers, the following error is threw:
```Python
---> 65     features = [self.polygonize(feat) for feat in self.data]
     66     [feat["properties"].pop("style") for feat in features]
     68     return {"type": "FeatureCollection", "features": features}

AttributeError: 'DrawControl' object has no attribute 'data'
```
Is it perhaps because this branch is depending on others?

dfguerrerom:
> > I'm not able now to select administrative layers, the following error is threw
> 
> You need to install a version of ipyleaflet > 0.14.0 to have access to data (changed in the seup.py but if you didn't reinstalled it you will face a bug)

Yeap, it was the version. Thanks.

sepal_ui/mapping/draw_control.py
```diff
@@ -51,3 +52,47 @@ def hide(self):
...
+    def to_json(self):
+        """
+        Return the content of the DrawCOntrol data without the styling properties and using a polygonized representation of circles.
+        The output is fully compatible with __geo_interface__.
+
+        Return:
+            (dict): the json representation of all the geometries draw on the map
+        """
+
+        features = [self.polygonize(feat) for feat in self.data]
```
dfguerrerom: with polygonize, we are mutating each json geometry from the original `DrawingControl`, and as in the next line we are removing the `style` property, and an error will be thrown if you call twice `to_json` method... I tried .`.copy` over the feat dict, but it didn't work, so I used `deepcopy`...

Apparently there is a bug in the `DrawingControl`, if you create more than circles, the firsts will lose their radius, so the `to_json` can't succeed...

12rambau: So as a starter, the feat should not be modified effectively nice catch and I hope the deepcopy do the trick. 
Which version of ipyleaflet are you using ? maybe we need to update to the latest one. Could you try that and let me know if the issue persist ?

dfguerrerom: Using ipyleaflet==0.16.0

dfguerrerom (approved):
I don't think the circle issue will cause problems in the short term. So I think we can merge it as it is.

PR #518, opened 2022-07-03
fix: use txt in mapbtn

### LINKED ISSUE #510: add posibility to add text in the map_btn
The current implementation of the map_btn only authorize to use logos. It would be nice to let the opportunity to use letters as in the SEPAL main framework (3 letters only in capital)

12rambau:
Fix #510


sepal_ui/mapping/map_btn.py
```diff
@@ -7,20 +7,27 @@
...
 
     logo = None
-    "(sw.Icon): a sw.Icon"
+    "(sw.Icon|str): the content of the btn"
 
-    def __init__(self, logo, **kwargs):
+    def __init__(self, logo=None, msg=None, **kwargs):
 
         # create the icon
-        self.logo = sw.Icon(small=True, children=[logo])
```
dfguerrerom, line 23: Can we receive "`icon`" key instead of logo? and later rename if necessary to whatever else?

12rambau: unfortunately we cannot use "icon" as it's a parameter of vuetify btn (https://vuetifyjs.com/en/api/v-btn/#props-icon) and we decided at some point not to override existing keys for something else. 

So I'm happy to change the name if  you have any other idea that is not part of the existing keys of the vuetify Btn component

dfguerrerom: Ok, we have to follow our own rules. We have replaced `type` by `type_` in some elements, what do you think about `icon_`?

12rambau:
I changed my initial idea, instead, there is 1 single argument `content` if it starts with "mdi" or "fas" it will be considered as an icon and if not it will be considered as a text. 
still compatible with previous implementation from master

PR #533, opened 2022-07-06
refactor style.py

### LINKED ISSUE #445: add TYPES to documentation
don't haw to deal with variables yet

12rambau:
Fix #445 

Ok so lot's of changes here. I was working on simply documenting the content of `styles.py` and realized that there was a real mix of many things in there: json, javascripts, css etc.. 

This PR is an attempt to rationalize the styles of sepal_ui: 

- predefined styles for layers and components are saved in the `json` folder 
- the custom ccs of all applications is saved in `css` folder 
- the js we use in some components (fullscreen, resizeTrigger etc...) goes to the `js` folder. 

these extra files have the correct extension making easier to sanitize.

Then to point to these folders I added 3 extra folder Path in styles.py. 

I also removed color from style as it's a duplicate from `sepal_ui.__init__`. The only reason it was initialized here was because colors in the "jsonlike" were also defined in `styles.py`. 

I extracted color definition from the json files (specifically for layers). If we want to move toward a theme change without restart, colors need to be defined where they are used to create callbacks.

Finally I did what was expected at the beginning, I wrote down the styles.py file documentation page.


sepal_ui/mapping/map_btn.py
```diff
@@ -22,10 +25,14 @@ def __init__(self, content, **kwargs):
...
 
+        # create the style from default
+        style = json.loads((ss.JSON_DIR / "map_btn.json").read_text())
+        style.update(background=color.bg)
+
         # some parameters are overloaded to match the map requirements
         kwargs["color"] = "text-color"
         kwargs["outlined"] = True
-        kwargs["style_"] = " ".join([f"{k}: {v};" for k, v in map_btn_style.items()])
+        kwargs["style_"] = " ".join([f"{k}: {v};" for k, v in style.items()])
```
dfguerrerom, line 35: Just a comment here. Wouldn't it be better if we set the colors of the button with the vuetify properties instead of changing the element CSS properties?

dfguerrerom (commented):
I haven't seen these changes before, and I want to submit a comment.
I understand the advantage of having the style files organized in different folders, especially `javascript` and `css`... what is not totally clear for me are the `JSON` files. 
I saw that we are updating some of the properties directly after calling the objects where are going to be used, to be sure they're adapted to the current theme (as here: https://github.com/12rambau/sepal_ui/blob/a6836ec2c4e25386d2167379f950a9fbe08fc4c0/sepal_ui/mapping/map_btn.py#L30).
So IMO I think that we are creating an extra step within the implementation. We have created a color Box to deal with the colors, wouldn't be useful do the same with the json files?

12rambau:
So I remember that there was an explanation related to the coloring of the maps. I add in mind this idea of using the `v.theme` trait to control dark and light themes on the fly but I needed this JSON. 

Let me move this question to an issue so that I don't forget and I'll find back an explanation/or change it

PR #535, opened 2022-07-06
create a function to check for never used keys

### LINKED ISSUE #420: create a translator function to check the use of the keys
If you are updating many time the same application you may end up removing some or all the existing keys. It complex to visually assess if all the remaining keys in the dict are used. 

Maybe a parser could be interesting to check all the folder files and validate the keys that are used. 

Usage will be of course for developer only

12rambau:
In project that have been maintained for a long time, I've lost track of which keys is used where, leading to sometime many keys in the translator that are never used anywhere. This new method is checking if the keys are called at least once in the files of the application. 

Only test the straight foward calling method but other pattern can be added in other PR.

I took advantage of having it running to check sepal_ui itself and removed the leys that are never used.

Fix #420


12rambau:
I think I'm  the only one using it, let's deploy it and see how it goes

PR #539, opened 2022-07-08
add the gdrive/gee methods

### LINKED ISSUE #16: add a gdrive module
Lots of the computation performed with python in sepal implies an interaction between gee (and gdrive) and the sepal folder. 

Do you think it could be useful to create standard drive_handler object ? (inspired by the one in pysmm and glad_alert) 

with some standard method like 
- get_items
- download to sepal
- print list
- get file id 
- delete file
...

If some application requires specific implementation you can just create your custom drive_handler inheriting from this one

12rambau:
Fix #16 

The idea is to add the needed methods to easily deal with exportation using the Gdrive account of the user as a proxy between GEE and Local folders. here is a first batch of methods, let me know if some are missing.


dfguerrerom:
I like this idea, what do you think if we create a GeeDrive (or similar) model object to deal with this kind of tasks? instead of separate scripts. We could then associate a view when it's needed or just use the model when we have other needs. This will be similar to the `planetapi` package. I could also help in this one if you don't mind.

12rambau:
> I like this idea, what do you think if we create a `GeeDrive `(or similar) model object to deal with this kind of tasks? instead of separate scripts. We could then associate a view when it's needed or just use the model when we have other needs. This will be similar to the `planetapi` package. I could also help in this one if you don't mind.

Sounds reasonable to me. I've leave this one on stale for a while so if you want to PR/push things help yourself !

dfguerrerom:
I just updated the `service` to access through SEPAL with latest versions... I haven't tested the other methods, still WIP.

dfguerrerom:
closed in favor of https://github.com/openforis/sepal_ui/pull/963 which implements GDriveSession

PR #540, opened 2022-07-08
refactor: move decorator to their own modules

### LINKED ISSUE #267: Move decorators to its own module
As the number of decorators is increasing, we should consider moving the decorators to their own module, and separate them from the scripts.

12rambau:
I was forced to use some tricks to maintain the function available in su. It will be fully removed in 3.0

Fix #267


12rambau:
the patch is down because of the duplication of init_ee, I'm ignoring it

PR #571, opened 2022-08-11
feat: add Sepal Map method to create and set legend

### LINKED ISSUE #570: add_legend feature
Now, as we got rid of the geemap library, one important and missing feature is `add_legend`, would be nice to get it back natively.

dfguerrerom:
The PR, contains the following:

- An own map legend method and class with ipyvuetify elements instead of ipywidget HTML as it was done by the previous `geemap` implementation.
- A new parameter and trait to switch between vertical or horizontal view
- The legend main argument is just a simple dictionary with label:color, which is also a trait, so it can be updated if required.

Before pushing the test I'd like to know if you have another thing/feat that we can add or remove.

WIP

map_legend

Fix #570

### DIFF
--- a/sepal_ui/mapping/legend.py
+++ b/sepal_ui/mapping/legend.py
@@ -1,0 +1,101 @@
+from ipywidgets import HTML
+from traitlets import Bool, Dict, Unicode, observe
+
+import sepal_ui.sepalwidgets as sw
+from sepal_ui.scripts import utils as su
+
+
+class Legend(sw.Card):
+    """
+    Custom Card to display and update map legend
+
+    """
+
+    title = Unicode("Legend").tag(sync=True)
+    "str: title of the legend. defaults to 'Legend'"
+
+    legend_dict = Dict({}).tag(sync=True)
+    "dict: dictionary with key as label name and value as color"
+
+    vertical = Bool(True).tag(sync=True)
+    "bool: whether to display the legend in a vertical or horizontal way"
+
+    _html_table = None
+
+    def __init__(self, legend_dict, title="Legend", vertical=True, **kwargs):
+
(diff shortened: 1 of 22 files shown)


12rambau:
The fact that we can now dynamically change the content is super useful ! 

Is there a title or is it always "legend" ?

12rambau (commented):
Is there any reason not to set Legend directly as a map widget ? (as we did with the others that live in mapping package)

dfguerrerom:
> Is there any reason not to set Legend directly as a map widget ? (as we did with the others that live in mapping package)

Well, I think because the legend is expecting one parameter to be instantiated, the `legend_dict`, otherwise I think there's no point to have a legend without nothing, what do you think?

sepal_ui/mapping/legend.py
```diff
@@ -0,0 +1,112 @@
...
+        self.set_legend(None)
+
+    def __len__(
+        self,
+    ):
+        """returns the number of elements in the legend"""
+        return len(self.legend_dict)
+
+    @observe("legend_dict", "vertical")
+    def set_legend(self, _):
```
12rambau, line 51: don't you think it could make sense to raise an error if the legend_dict is empty ?

dfguerrerom:
I think it was better to hide the control if there is no dictionary, so we can handle the case when someone instantiates the legend without any parameter but is planning to add it later...
Something like this:
```Python

from sepal_ui.mapping import LegendControl

legend_control = LegendControl()
map_.add_control(legend_control)

legend_control.legend_dict = legend_dict
```

What do you think?

12rambau (approved):
Excellent work, I really like the look and feel of it. 

The errors are coming from the documentation as some files have changed from the sepal-doc side (it's already solved in `main`)

Issue #519, opened 2022-07-03
planet API is making the CI crash


12rambau:
All the rececent CI based on master are crashing as new deprecation warning are raised during the Jupyter-sphinx execution: 

> WARNING: Cell printed to stderr:
[21](https://github.com/12rambau/sepal_ui/runs/7168163180?check_suite_focus=true#step:9:22)
/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/planet/api/__init__.py:38: ClientV1DeprecationWarning: The planet.api module is deprecated and will be removed in version 2.0.0. For more details please see the discussion at https://github.com/planetlabs/planet-client-python/discussions.

I think we need to have a look to see if we need to fix planet version

12rambau:
The solution might be here: https://github.com/jupyter/jupyter-sphinx/issues/178

dfguerrerom:
Anyways I think we have to move to version 2. v1 is ugly.

12rambau:
I made sure sphinx stop throwing warning to our face, we'll move to v2 whenever we want

PR #578, opened 2022-08-24
Planet upgrade #577

### LINKED ISSUE #577: Upgrade to planet v2
As pointed out by #519 and to be aligned with the latest developments by `planet-client-python` library, we should migrate our models to the latest version: planet 2.0.

### LINKED ISSUE #579: StateIcon tooltip is fixed after trait changes
When using the `StateIcon` and changing its state accordingly to its states, the new value for the tooltip keeps fixed until you pass the mouse over the icon, if you don't pass the cursor, it will keep displayed forever. I think this is a bug from the tooltip and slots from ipyvuetify, but I will keep exploring. In the meanwhile, I think we can remove the tooltip of the `StateIcon`

dfguerrerom:
Related with #519 and #577.

I've updated the model, view, and test of planetapi. However, I think I will remain it as a draft until they officially launch version 2.0, which apparently will be before the end of the year, or I'll keep testing so we can use a pre-release and move forward.

The pre-release version 2.0a2 is already available to use https://github.com/planetlabs/planet-client-python/tree/2.0a2. This new version is more robust than the previous one and it allows us to create asynchronous tasks, super useful if we are planning to do some multi-processes requests. 

The `PlanetModel` wrapper is a high-level model to easily manage the `async` calls and conserve the same structure that the other modules as `planet_active_fires_explorer`  are using. As previously, we can keep developing this model to be the generic entry point of the `sepal-contrib` modules.

#### update 26/08:

- I think we can use the prerelease planet v2 version. Our queries are quite simple, so I don't see any problem by using this one.

- Regarding the https://github.com/sepal-contrib/alert_module/issues/56, I've added the logic to query from the active subscriptions all the nicfi levels.

planet_subscriptions

Fix #577 
Fix #579


dfguerrerom:
The error is related to the `Translator`, I didn't modify any of these files here, so I think that's to be checked in another PR. 
IMO, this branch is ready to be reviewed and tested, we can use the planet 2.0 pre-release.

12rambau (commented):
let's say I have more than 3 subscription, would it be possible to add a "other" section of subscription and make sure that the connection is valid ?

dfguerrerom:
> let's say I have more than 3 subscription, would it be possible to add a "other" section of subscription and make sure that the connection is valid ?

I've added support to "others" subscriptions, and I've merged `level` buttons into just one. I think you can give it a try the view. The only thing that I'm not so sure about is the height of the component since now is increasing accordingly with the number of subscriptions in each category.  
If we agree on this, I'll test the view component.

12rambau (commented):
I like the new look a lot ! 

2 comments: 
- I think you should add a small left margin on the "from" section to align it to the title 
- I think you should set the max height to 1.2x the height of 1 card, so that if there are multiple card the user will see that he needs to scroll
Capture d’écran 2022-08-30 à 19 30 50

12rambau (approved):
Everything's green from my side! 
Merge it whenever you want

PR #581, opened 2022-08-29
Legend suggestion

12rambau:
small implementation change: 

`Lengend` is now not a `Card` but a `WidgetControl`. It allows more flexibility to add it to the map: 

people that are used to ipyleaflet: 
```python 
legend = sm.LengendControl({})
map = SepalMap()
map.add_control(legend)
```

people that know the lib very well 
```python 
map = SepalMap()
map.add_legend({}, position="topleft")
```

I think the implementation you did pave the way for all the controls in general: 
- the ipyleaflet way (created outside used outside) 
- the semi auto (created with a specific method, added to the member)
- the lazy one (create at init state, added to the members)

This will of course be handled in another PR and another issue

### DIFF
--- a/sepal_ui/mapping/legend_control.py
+++ b/sepal_ui/mapping/legend_control.py
@@ -1,3 +1,4 @@
+from ipyleaflet import WidgetControl
 from ipywidgets import HTML
 from traitlets import Bool, Dict, Unicode, observe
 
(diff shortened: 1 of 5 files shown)


sepal_ui/mapping/legend_control.py
```diff
@@ -6,53 +7,75 @@
...
         self.legend_dict = legend_dict
         self.vertical = vertical
 
-        self.html_title = sw.Html(tag="h4", children=[f"{self.title}"])
+        # generate the content based on the init options
+        self._html_title = sw.Html(tag="h4", children=[f"{self.title}"])
         self._html_table = sw.Html(tag="table", children=[])
-        self.children = [self.html_title, self._html_table]
+
+        # create a card insode the widget
```
12rambau, line 48:
```suggestion
        # create a card inside the widget
```

dfguerrerom (approved):

PR #590, opened 2022-09-08
Set default style to vector layers in GEE

dfguerrerom:
As mentioned in #425, gee vector layers were not using the default style, with this PR, we will have this as a default behavior.

### DIFF
--- a/sepal_ui/mapping/sepal_map.py
+++ b/sepal_ui/mapping/sepal_map.py
@@ -27,7 +27,7 @@ from matplotlib import colorbar
 from matplotlib import colors as mpc
 from rasterio.crs import CRS
 
-from sepal_ui import color
+from sepal_ui import color as scolors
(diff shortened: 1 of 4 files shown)


sepal_ui/mapping/sepal_map.py
```diff
@@ -610,17 +610,31 @@ def add_ee_layer(
...
-            )
+
+            try:
+                image_fill = features.style(**vis_params).updateMask(const_image)
+                image_outline = features.style(**vis_params)
+
+            except AttributeError:
+                # Raise a more understandable error
+                raise AttributeError(
+                    "You only can use the following styles: 'color', 'pointSize', "
```
12rambau, line 634:
```suggestion
                    "You can only use the following styles: 'color', 'pointSize', "
```

12rambau (approved):
Appart from the small typo it seems perfect to me.

PR #591, opened 2022-09-09
fix: remove alert from aoi_model and add model as optional in aoi_view

### LINKED ISSUE #589: AoiView doesn't allow start with custom model and model requires alert
### AoiView
In one of my apps, my approach is to create the main model which will handle and receive other models, something like this:

```python

aoi_model = AoiModel()
planet_model = PlanetModel()

my_model= MyModel(aoi_model, planet_model)

#...

# and then...

aoi_view = AoiView(model=aoi_model)

```
Having this I'll be able to store all the inputs from the models in one single place, and so, perform all the processes I would like to run, however, with the current way as the `AoiView` is created, I can't do that in a simple way, so I have opted for hard-copy the full view in my workflow, which is not convenient.

It would be just a matter of making it an input parameter in the view and creating an operator to use it if is set.

### AoiModel

After hard-copying `AoiView` and trying to instantiate the `AoiModel`from the outside, I get this error...

```python
TypeError: __init__() missing 1 required positional argument: 'alert'
```

This is because we are requiring an `Alert` to process the user outputs from within the model. I think we should get rid of `view` elements in the `model` and try to manage them with our `loading_button` decorator to catch errors and warnings, and perhaps create the successful outputs directly in the view... OR... try to improve the decorator and capture the successful messages from the model in the `view alert` (something like capturing `prints` or some new `Message` class).

I say this because I think the model has to be totally independent of the view, and even be able to display its outputs in two ways... as pure python and view messages.

@12rambau what do you think? I'm not sure if I was clear on my idea.

dfguerrerom:
This is a solution in order to close #589

### DIFF
--- a/sepal_ui/aoi/aoi_model.py
+++ b/sepal_ui/aoi/aoi_model.py
@@ -6,6 +6,7 @@ from urllib.request import urlretrieve
 import ee
 import geopandas as gpd
 import pandas as pd
+from deprecated.sphinx import deprecated
 from ipyleaflet import GeoJSON
 from traitlets import Any
 
@@ -123,9 +124,6 @@ class AoiModel(Model):
     folder = None
     "str: the folder name used in GEE related component, mainly used for debugging"
 
-    alert = None
-    "sepal_ui.sepalwidgets.Alert: the alert to display outputs"
-
     default_vector = None
     "(str|pathlib.path: the default vector file that will be used to produce the gdf. need to be readable by fiona and/or GDAL/OGR"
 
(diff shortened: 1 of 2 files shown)


12rambau (approved):
LGTM! 

let's keep the alert member for a little while as many of the current apps have been built using it. I would love to have a smooth transition

dfguerrerom:
> LGTM!
> 
> let's keep the alert member for a little while as many of the current apps have been built using it. I would love to have a smooth transition

Sure, I left the alert member as optional...

PR #601, opened 2022-10-13
feat: improve date picker widget customization

### LINKED ISSUE #600: Datepicker is not fully customizable
As our main `DatePicker` usage is as in its "menu" form, it is not handy to set some use cases:
- set a min_, max_ value directly (you have to `datepicker.children.....min_`...)
- set a default initial value with `v_model` since it is hardcoded from the beginning
- the `jslink` "link" will only work if the change is made from a "js" event, but not if you want to link the values since the initialization.

dfguerrerom:
closes #600

### DIFF
--- a/sepal_ui/sepalwidgets/inputs.py
+++ b/sepal_ui/sepalwidgets/inputs.py
@@ -6,6 +6,7 @@ import ee
 import geopandas as gpd
 import ipyvuetify as v
 import pandas as pd
+from deprecated.sphinx import versionadded
 from ipywidgets import jslink
 from natsort import humansorted
 from traitlets import Any, Bool, Dict, Int, List, Unicode, link, observe
@@ -30,6 +31,10 @@ __all__ = [
 ]
 
 
+@versionadded(
+    version="2.13.0",
+    reason="Empty v_model will be treated as empty string: :code:`v_model=''`.",
+)
 class DatePicker(sw.Layout):
     """
     Custom input widget to provide a reusable DatePicker. It allows to choose date as a string in the following format YYYY-MM-DD
--- a/sepal_ui/sepalwidgets/inputs.py
+++ b/sepal_ui/sepalwidgets/inputs.py
@@ -11,7 +11,6 @@ from ipywidgets import jslink
 from natsort import humansorted
 from traitlets import Any, Bool, Dict, Int, List, Unicode, link, observe
 
-import sepal_ui.sepalwidgets as sw
 from sepal_ui import color
 from sepal_ui.frontend import styles as ss
 from sepal_ui.message import ms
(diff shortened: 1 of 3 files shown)


sepal_ui/sepalwidgets/inputs.py
```diff
@@ -10,6 +10,7 @@
 from natsort import humansorted
 from traitlets import Any, Bool, Dict, Int, List, Unicode, link, observe
 
+import sepal_ui.sepalwidgets as sw
```
12rambau, line 13: As we are living in the sepalwidgets module you cannot yet import it inside one of its files (as __ini__.py is also importing inputs and all the others). In short, that's a circular reference and it can lead to an undebugable situation (that we already face at the beginning). 
Could you fall back to the previous implementation (with v.Layout and SepalWidget class) ?

sepal_ui/sepalwidgets/inputs.py
```diff
@@ -102,7 +106,7 @@ def check_date(self, change):
         self.date_text.error_messages = None
 
         # exit immediately if nothing is set
-        if change["new"] is None:
+        if not change["new"]:
```
12rambau, line 109: I try my best not to use `not` like this anymore as `""`, `False` and `None` will all be treated the same way. Is there any reason why you removed the default value of the textField ?

12rambau (commented):
I think you coded this at the very beginning of the lib and the difference between "link" and "jslink" was still unclear for us. That's clearly a nice refreshment.
 I have few comments though:

sepal_ui/sepalwidgets/inputs.py
dfguerrerom: ok, so for this very first point, I use to bind the `DatePicker` `v_model` with a `traitlets.Unicode` in the model, Unicode because it is what the widget returns, if we use `None` as the default "`v_model`" it will complain the trait link (we can create the `Unicode(allow_none=True)` tho), so, with this change I —this is just my personal preference— just don't like the `change["new"] == ""`, that's why I've changed it.

And for the second point, you  mean the `v_model` from the `TextField`? I've removed it because it will be linked by chain events by the `link`, so I don't think it is fully neccessary.
```python
        link((self.date_picker, "v_model"), (self.date_text, "v_model"))
```

what do you think?.

12rambau: I agree with making full use of the traitlets and the type checking so big no for the `allow_none=True` I dived into the PEP 8 recommendation and apprently: 

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

So as the Unicode cannot be anything else than a string that's legit. 

Could you add a [`versionchanged`](https://pypi.org/project/Deprecated/) to the class to notify this change (I'm positive I'm checking `is None` in some modules) ?

12rambau (approved):

12rambau:
you just need to update the tests as the kwargs are not passed to the Layout anymore

Issue #602, opened 2022-10-14
Select widget items not displayed on full screen when embedded on expansion panel


dfguerrerom:
**Describe the bug**
Similar as https://github.com/widgetti/ipyvuetify/issues/141, `Select` items are not displayed on the latest `FullScreen` implementation of `sepal_ui` when the element is embedded on an `ExpansionPanel`.

**To Reproduce**

```python
from ipyleaflet import WidgetControl
import ipyvuetify as v
from sepal_ui.mapping import SepalMap, FullScreenControl

expansion = v.ExpansionPanels(
    children=[
        v.ExpansionPanel(
            children=[
                v.ExpansionPanelHeader(children=["header"]),
                v.ExpansionPanelContent(children=[
                    v.Select(items=[f"{i}" for i in range(10)])
                ])
            ]
        )
    ]
)

map_=SepalMap()
map_.add_control(
    FullScreenControl(map_, position="topleft", fullscreen=True, fullapp=True)
)
control = WidgetControl(widget=expansion)
map_.add_control(control)
map_
```

**Expected behavior**
Just display items as you would expect without full screen

PR #604, opened 2022-10-14
Fix z-index to menu content

dfguerrerom:
This aims to close the issue #602 . Basically, it just increases the `.v-menu__content` zindex.


12rambau (approved):
Nice catch

dfguerrerom:
The errors we are having apparently come from GADM...

PR #607, opened 2022-11-17
fix: keep default asset when reloading

### LINKED ISSUE #596: custom assets are removed when reloading `AssetSelect` component
**Describe the bug**
After reloading the `AssetSelect` widget, the custom assets that were created in the definition are removed from the list.

**To Reproduce**
1. Create an `AssetSelect` component with a custom asset
2. use the reload button
3. check the disappeared custom asset(s)

**Expected behavior**
Custom assets should be kept even if user has selected or used the reload button.

12rambau:
Fix #596 

I simply moved the `default_asset` addition into the `_get_items` method. 
I also trigger the full `_get_items` when the `default_asset` is changed

```python 

widget = sw.AssetSelect(default_asset=["users/bornToBeAlive/random300"])
widget

widget.default_asset = ["users/bornToBeAlive/random150"]
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
     hooks:
       - id: black
         stages: [commit]
-  - repo: 'https://gitlab.com/pycqa/flake8'
+  - repo: 'https://github.com/pycqa/flake8'
     rev: 3.9.2
     hooks:
       - id: flake8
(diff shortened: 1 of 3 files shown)


12rambau:
@dfguerrerom just ignore the codecov patch value, `input.py` is covered at 100% there is nothing I can add.

dfguerrerom (approved):
simplier!

Issue #630, opened 2022-11-27
improve the testing process using pytest plugin 


12rambau:
need to decide which from this list are useful in our case:
https://towardsdatascience.com/pytest-plugins-to-love-%EF%B8%8F-9c71635fbe22

12rambau:
- [x] pytest-sugar
- [ ] pytest-icdiff
- [ ] pytest-cov ?
- [ ] [pytest-instafail](https://pypi.org/project/pytest-instafail/)
- [ ] [pytest-timeout](https://pypi.org/project/pytest-timeout/)
- [ ] [pytest-deadfixtures](https://pypi.org/project/pytest-deadfixtures/)

Issue #626, opened 2022-11-25
refactor add_layer and remove_layer


12rambau:
In the latest release of `ipyleaflet`, all components (controls, layers etc...) are added to the map using `add` instead of `add_layer` and `add_control`. 

In the lib we are overwriting the `add_layer` in multiple ways. The implementation does not rise DeprecationWarning anymore but we should consider refactoring the way we add layers to the map so that newcomers don't get lost.

Issue #621, opened 2022-11-22
isort is breaking our CD/CI


12rambau:
first report in https://github.com/12rambau/sepal_ui/actions/runs/3525494576/jobs/5912282242, I opened an issue on their repository (https://github.com/PyCQA/isort/issues/2006), I'm pretty sure I'm missing a config parameter.

I created the `isort` branch to experiment on it.

Issue #562, opened 2022-07-25
allow other values for progress


12rambau:
Now that we are supporting tqdm it should be possible to support progress values that are not between 0 and 1. https://github.com/12rambau/sepal_ui/blob/c15a83dc6c92d076e6932afab4e4b2987585894b/sepal_ui/sepalwidgets/alert.py#L98

PR #635, opened 2022-11-28
test: update the test suit

12rambau:
## context 

It all started when I realized 2 things: 
1. the earthengine tests are not run in the github CD/CI 
2. the earthengine tests are all relying on my account and a sepal_ui folder within it

## objectives

The main objectives of this huge PR are:
- make sure that all the tests are run in the test suit whatever the environment 
- run the test on local earthengine account without the support of my `sepal-ui` folder 
- increase test speed 
- improve the display of the unit tests

## content 

In the end I did way much more things than expected, I'll try to be as exhaustive as possible to make the review easier: 
- following [these recomandation](https://github.com/giswqs/geemap/discussions/1341) I set up an `EARHENGINE_TOKEN` in the Github action secrets and make sure `init_ee` could pick it up. 
- refactor `conftest.py` so that it creates all the needed assets in a specific folder
- use a hex `_hash` for all the assets so that tests can be run in parallel. if not some names would be conflicting everywhere.
- update the tests so that we check for `ee.credentials` to skip gee-related tests
- split all mixed tests in `aoi_model` and `reclassify_model` so that gee and local behavior are tested separately
- make good use of the scope of the pytest fixture to speed up file creation (by avoiding duplication)
- as recommended in this [article](https://towardsdatascience.com/pytest-plugins-to-love-%EF%B8%8F-9c71635fbe22) I added `pytest-sugar`, `pytest-icdiff` and `pytest-instafail` (only used in the github workflow)
- remove nearly all calls to deprecated methods from our lib and other (I opened some new issues about that while making the modification e.g. leaflet `add` method)
- global speed boost
- create a `test_decorator.py` file
- update the contribution section with the earth engine ENV setup

## side notes

- The linting test will fail until we find a solution for this issue: https://github.com/PyCQA/isort/issues/2006
- the coverage is decreasing because I stopped testing deprecated methods without removing them (ignore codecov failure for now)

# related issues

they are never completely solved but that's a good introduction 
#562 
#621 
#626 
#630 

## conclusion

Happy to hear feedback but as it's not changing any of the lib behavior I'll try to merge it as soon as possible to start follow-up PR

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -25,7 +25,6 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      max-parallel: 1
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
 
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -72,7 +72,7 @@ jobs:
           assert len(unexpected) == 0
       
       - name: test with pytest
-        run: coverage run -m pytest --color=yes tests
+        run: coverage run -m pytest --color=yes --instafail tests
         
       - name: build the template panel application 
         if: matrix.python-version == '3.8'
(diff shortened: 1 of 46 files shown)


dfguerrerom (approved):
:)

PR #645, opened 2022-11-29
Fontawesome V6

### LINKED ISSUE #576: Change to free tagged icons v.5 on fontawesome?
Our File Input menu now is not displaying most of the different icons. I think we have to decide if we are going to have a paid version (and wait) or keep using v.5 but with free tagged icons to avoid this issue. What do you think?

image

12rambau:
- use fontawesome V6 
- change all the `fas` and `far` occurences by `fa-solid`(res. `fa-regular`)
- change the icons that were not free to their free alternative

**note**
`fas` will continue to work in v6 for quite some time so application will not need to change everything but their sizing is a bit different. 

Fix #576


12rambau:
I need this PR to be merged for the conda release. I'm overpassing review checks. Don't hesitate to reach in the issues if needed

PR #646, opened 2022-11-29
fix: update_progress accept values >1

### LINKED ISSUE #562: allow other values for progress

12rambau:
- by default ùpdate_progress continues to refuse values >1 (no regression)
- the user can select a different `total` parameter for tqdm and then the value should only be <`total`

Fix #562

### DIFF
--- a/sepal_ui/sepalwidgets/alert.py
+++ b/sepal_ui/sepalwidgets/alert.py
@@ -94,9 +94,10 @@ class Alert(v.Alert, SepalWidget):
         self.show()
 
         # cast the progress to float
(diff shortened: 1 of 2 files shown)


12rambau:
I need this PR to be merged for the conda release. I'm overpassing review checks. Don't hesitate to reach in the issues if needed

PR #708, opened 2023-01-23
add a custom zoomControl

12rambau:
In the current implementation, the zoom control is a normal ipyleaflet widget so its display is kinda tricky: 
- transparent or not depending on the execution 
- use the voila background 
- don't respect theming 
- don't use the theme colors

This is a custom implementation of a python zooming widget. It is using the same look & feel as the other buttons (Vuetify material) and respect the themes. 

I think I cover all the potential situations.

Capture d’écran 2023-01-23 à 12 14 33

### DIFF
--- a/sepal_ui/mapping/zoom_control.py
+++ b/sepal_ui/mapping/zoom_control.py
@@ -57,7 +57,7 @@ class ZoomControl(WidgetControl):
     def zoom(self, widget: MapBtn, *args) -> None:
         """update the zoom according to the clicked btn."""
         # read min and max zoom
-        max_zoom = self.m.max_zoom or 20
+        max_zoom = self.m.max_zoom or 24
         min_zoom = self.m.min_zoom or 0
 
         # computed zoom
--- a/sepal_ui/mapping/zoom_control.py
+++ b/sepal_ui/mapping/zoom_control.py
@@ -1,0 +1,69 @@
+"""
+Customized ``Control`` to zoom in and out on the map.
+"""
+from typing import Optional
+
+from ipyleaflet import Map, WidgetControl
+
+from sepal_ui import sepalwidgets as sw
+from sepal_ui.mapping.map_btn import MapBtn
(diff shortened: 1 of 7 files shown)


sepal_ui/mapping/zoom_control.py
```diff
@@ -0,0 +1,69 @@
...
+        super().__init__(**kwargs)
+
+        # add js behaviour
+        self.plus.on_event("click", self.zoom)
+        self.minus.on_event("click", self.zoom)
+
+    def zoom(self, widget: MapBtn, *args) -> None:
+        """update the zoom according to the clicked btn."""
+        # read min and max zoom
+        max_zoom = self.m.max_zoom or 20
```
12rambau, line 60:
```suggestion
        max_zoom = self.m.max_zoom or 24
```

I need to check but I think, 24 is actually the maximum default

sepal_ui/mapping/zoom_control.py
```diff
@@ -0,0 +1,69 @@
...
+        Args:
+            m: the map to manipulate
+            kwargs: any ``ipyleaflet.widgetControl`` arguments
+        """
+        # init the map
+        self.m = m
+
+        # create the 2 btns
+        self.plus = MapBtn("fa-solid fa-plus", _metadata={"step": 1})
+        self.minus = MapBtn("fa-solid fa-minus", _metadata={"step": -1})
```
dfguerrerom, line 38: at some point you recommended use `attributes` parameter instead of `_metadata`... since attributes will always be a dictionary, in this case it's not harmful, but for consistency could we change it to attributes? and if not, in which cases you'd prefer one over the other?

dfguerrerom (commented):
Just a comment

sepal_ui/mapping/zoom_control.py
12rambau: sure thing

PR #710, opened 2023-01-24
replace the `LayerControl` to modify visibility

12rambau:
Not completed yet but I would like to have your feedback before writing the tests.

The main idea is that the buit-in `ipyleaflet.LayerControl` don't allow to edit the opacity on the fly. It's a first step before adding the streching options. 

what has changed: 
- create a `SimpleSlider` widget. It took me some time but I find how to align sliders. you need to use some small CSS tricks that are automatically setup if you use this one instead of a `Slider`
- override Radio. Radio cannot be used as stand-alone objects and they have no trait to tell if they are currently activated or not. I created a traitlet overlay on the RadioGroup/Radio so that we can pick up the state of a radio button ... from the radio button 
- with the 2 previous helper I built a menuControl that display the layers as they 2 lists: the basemaps (only one at a time) and the other (checkbox). You can change the opacity using the slider and also edit the visibility by checking/unchecking. The widget is bidirectional, meaning that if you manually change the visibility of a layer the widget is updated. The content is regerated every time the `layers` trait of the map changes. 

here is some code to test it: 
```python 
from sepal_ui import mapping as sm
import ee 

ee.Initialize() 

layer_1 = ee.Image("users/wiell/forum/visualization_example")
m = sm.SepalMap(["HYBRID", "CartoDB.Positron"], vinspector=True)
viz = m.get_viz_params(layer_1)
for i in viz:
    viz_name = viz[i]["name"]
    m.add_ee_layer(layer_1,name=f"layer {viz_name}", viz_name=viz_name)
m.zoom_ee_object(layer_1)
m.layers[2].opacity = .5
m
```

and the result : 
Enregistrement de l’écran 2023-01-24 à 23 39 49

@dfguerrerom what do you think ?


12rambau:
first comment from my side: 
- this menu control should be independent from the others, meaning that opening it should not close the others.

sepal_ui/sepalwidgets/radio.py
```diff
@@ -0,0 +1,67 @@
...
+
+        Change the v_model of every subsequent radios buttons according to new value of the radioGroup.
+        children that are not Radios will be ignored.
+        """
+        for w in self.search_radios(self):
+            w.active = w.value == self.v_model
+
+        return
+
+    def search_radios(
```
dfguerrerom, line 54: What about if we use the `get_children` native SepalWidget method? it will just matter to add an "id" to radios.

sepal_ui/mapping/layers_control.py
```diff
@@ -0,0 +1,169 @@
...
+from ipyleaflet import Map, TileLayer
+from ipywidgets import link
+
+from sepal_ui import color
+from sepal_ui import sepalwidgets as sw
+from sepal_ui.frontend import styles as ss
+from sepal_ui.mapping.menu_control import MenuControl
+
+
+class BaseLine(sw.Html):
```
dfguerrerom, line 17: What about changing the name to `BaseRow` and `LayerRow`

dfguerrerom (commented):
First of all, it looks great, congrats!
Some comments from my side...
- I know we are following the style of the previous `LayerControl`, but I thing it would be nice if we reverse the order, so top layers in the map are also top in the list, in the same way as all the GIS software, what do you think?
- something minor, but when you hold the slider tip to the end, an x-scroll bar appears, could we force it to hide?
- the transparency will be something mandatory, the `LayerControl` will always display that option, right? if so, what do you think if we change the order of the sliders and checkers?, sliders to the left and checkers and name to the right column? in this way if you're not interested on change the opacity, you don't have to move the cursor farther, the checkers will be closer to the cursor, what do you think? does it make sense?

sepal_ui/sepalwidgets/radio.py
12rambau: Good point ! 
maybe I could update `get_children` to make it more customizable (like select the property and the value we return a list) but I need to think about it to avoid no-regression

12rambau:
> I know we are following the style of the previous LayerControl, but I thing it would be nice if we reverse the order, so top layers in the map are also top in the list, in the same way as all the GIS software, what do you think?

agreed but I will keep the grouping (base, layers) whatever the order, radio button need to stay close to each other

12rambau:
I did some updates: 

- update class names: `HeaderRow`, `BaseRow`, `LayerRow`
- implement a `group` feature for menu control. If they are in the same group they cannot be displayed at the same time, if not, nothing happens. `LayerControl` is added to `SepalMap` with `group=-1`. It should interact with nothing.
- implement a new `get_children`. Richer with `klass` and `attr` option. It continues the support for the old implementation (`_id`). I just decided to drop the transformation into single elements. It avoids the nested `def` and mimics the Html search (`getElementsById`, `getElementByTagName` ... etc)
- review the display: single table with headers, change buttons colors, trigger on hover (which prevents the x-sliding), order of the layers, right align the btns, add hovering on line to avoid confusion, use extra empty row to avoid cut of the last circle hover

preview :

dfguerrerom:
That looks amazing!

PR #711, opened 2023-01-25
fix: read resize_trigger js as text. Closes #709

### LINKED ISSUE #709: Resize event is not triggered
I was working on some PySMM issues and I came across the resize trigger issue. The JavaScript snippet doesn't run. Could it be related to the fact that the template is not part of the instance?

https://github.com/12rambau/sepal_ui/blob/5727303e2a0e853236b6028c58067f5dd680ff38/sepal_ui/frontend/resize_trigger.py#L23

dfguerrerom:
It was only because didn't read the js file.


12rambau (approved):
Of course ! 

nice catch I'm merging right now

PR #720, opened 2023-01-31
make the aoi style customizable

### LINKED ISSUE #706: make the color of the aoiView created object controlable
At the moment changing the color displayed by the AoiView once an AOI have been computed is very heavy and require to override _update_aoi method. 

This value should be exposed to the end user.

12rambau:
Fix #706 

As we now use the AOIView more often in custom map style application it became critical to modify the rendering of the AOI on the map natively. 

With this PR, the developer can change the AOI display directly when he creates the AOIView/Tile. as in the following example: 

```python
from sepal_ui import aoi 

aoi.AoiTile(map_style = {'opacity': 1, 'dashArray': '9', 'fillOpacity': 0.1, 'weight': 1})
```
Capture d’écran 2023-01-31 à 14 43 35

### DIFF
--- a/sepal_ui/aoi/aoi_model.py
+++ b/sepal_ui/aoi/aoi_model.py
@@ -599,9 +599,12 @@ class AoiModel(Model):
 
         return self
 
-    def get_ipygeojson(self) -> GeoJSON:
+    def get_ipygeojson(self, style: Optional[dict] = None) -> GeoJSON:
         """
         Converts current geopandas object into ipyleaflet GeoJSON.
+
+        Args:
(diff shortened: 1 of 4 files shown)


sepal_ui/aoi/aoi_model.py
```diff
@@ -616,8 +619,9 @@ def get_ipygeojson(self) -> GeoJSON:
             f["properties"]["name"] = self.name
 
         # adapt the style to the theme
-        style = json.loads((ss.JSON_DIR / "aoi.json").read_text())
-        style.update(color=color.success, fillColor=color.success)
+        if style is None:
+            style = json.loads((ss.JSON_DIR / "aoi.json").read_text())
+            style.update(color=color.success, fillColor=color.success)
```
dfguerrerom, line 624: we set as default color, `color.primary` when adding a layer: https://github.com/12rambau/sepal_ui/blob/aac8985731a4db13acdebfd34150482550a3db53/sepal_ui/mapping/sepal_map.py#L790, in this way, the `Aoi layer` will replace the default one?
In my opinion orange is prettier than green over dark theme.

PR #738, opened 2023-02-07
fix: change aoi color to primary

### LINKED ISSUE #736: set default AOI color to primary
we set as default color, `color.primary` when adding a layer: https://github.com/12rambau/sepal_ui/blob/aac8985731a4db13acdebfd34150482550a3db53/sepal_ui/mapping/sepal_map.py#L790, 
in this way, the `Aoi layer` will replace the default one?
In my opinion orange is prettier than green over dark theme.

_Originally posted by @dfguerrerom in https://github.com/12rambau/sepal_ui/pull/720#discussion_r1098104396_

12rambau:
Fix #736

### DIFF
--- a/sepal_ui/aoi/aoi_model.py
+++ b/sepal_ui/aoi/aoi_model.py
@@ -621,7 +621,7 @@ class AoiModel(Model):
         # adapt the style to the theme
         if style is None:
             style = json.loads((ss.JSON_DIR / "aoi.json").read_text())
-            style.update(color=color.success, fillColor=color.success)
+            style.update(color=color.primary, fillColor=color.primary)
 
         # create a GeoJSON object
         # attribution="SEPAL(c)" is not recognized yet


sepal_ui/aoi/aoi_model.py
```diff
@@ -621,7 +621,7 @@ def get_ipygeojson(self, style: Optional[dict] = None) -> GeoJSON:
         # adapt the style to the theme
         if style is None:
             style = json.loads((ss.JSON_DIR / "aoi.json").read_text())
-            style.update(color=color.success, fillColor=color.success)
+            style.update(color=color.primary, fillColor=color.primary)
```
dfguerrerom, line 624: but isn't it redundant? `add_layer` adds a default style when `GeoJSON` doesn't have one

dfguerrerom (approved):

PR #741, opened 2023-02-07
fix: update datepicker

### LINKED ISSUE #730: datepicker coded value changed are not rendered
If you set up a datepicker value from the code, the value is effectively stored in the datepicker but not displayed to the user: 

```python

toto = sw.DatePicker()
toto.v_model = '2023-01-06'
toto
```

### LINKED ISSUE #746: DatePicker renders duplicated when multiple created
After `sepal_ui>=2.13.0` when we implemented `layout_kwargs` attribute, an strange behavior started to appear: when multiple `DatePicker` are created and displayed in the same component, in somehow they start sharing its attributes, weird.

Try the following:

```python
import sepal_ui.sepalwidgets as sw


class DateSelector(sw.Layout):
    
    def __init__(self, *args, **kwargs):
        
        super().__init__(*args, **kwargs)
        
        self.w_unique_date = sw.DatePicker(label="Date")
        self.w_ini_date = sw.DatePicker(label="Start date")

        self.children = [
            self.w_unique_date,
            self.w_ini_date
        ]

date_selector = DateSelector()
date_selector
```

If you click over one of the date pickers, the other will be displayed.

image

After a lot of tests, I isolated the problem, but I cannot explain the reason of this... the problem can be solved in several ways:

- remove `layout_kwargs` from the `__init__` method.
- reassign `layout_kwargs={]` within the __init__ method.
- create a DatePicker as: DatePicker(layout_kwargs={})

Method 1 and 2 would drop the `layout_kwargs` purpose: customize the layout from the initialization of the instance.
Method 2 is horrible, I won't opt for this.

First thing, I would like to have an explanation for this, perhaps if we nail the problem we could solve it and keep layout customization support, otherwise, I would drop it.

What do you think @12rambau, do you want to give a try and confirm you have the same issue? and if you can try to give an explanation?

12rambau:
Fix #730, Fix #746

The `layout_kwargs` was init to `{}` and then updated with `setdefaults` for all `DatePicker` without option. That's a mutable object rookie mistake, they were all sharing the same. I made it optional and if not present, it's inited in the `__init__` method. 

can be tested by running: 

```python 

toto = sw.DatePicker()
toto.v_model = "2022-06-15"
toto
```

or for the duplication issue: 
```python 
import sepal_ui.sepalwidgets as sw


class DateSelector(sw.Layout):
    
    def __init__(self, *args, **kwargs):
        
        super().__init__(*args, **kwargs)
        
        self.w_unique_date = sw.DatePicker(label="Date")
        self.w_ini_date = sw.DatePicker(label="Start date")

        self.children = [
            self.w_unique_date,
            self.w_ini_date
        ]

date_selector = DateSelector()
date_selector
```

### DIFF
--- a/sepal_ui/aoi/aoi_view.py
+++ b/sepal_ui/aoi/aoi_view.py
@@ -350,7 +350,7 @@ class AoiView(sw.Card):
             self.aoi_dc.hide()
 
         # add a validation btn
-        self.btn = sw.Btn(ms.aoi_sel.btn)
+        self.btn = sw.Btn(msg=ms.aoi_sel.btn)
 
         # create the widget
         self.children = (
--- a/sepal_ui/sepalwidgets/btn.py
+++ b/sepal_ui/sepalwidgets/btn.py
@@ -6,9 +6,9 @@ All the content of this modules is included in the parent ``sepal_ui.sepalwidget
 
 Example:
     .. jupyter-execute::
-    
+
         from sepal_ui import sepalwidgets as sw
-        
+
         sw.Btn()
 """
 
@@ -74,8 +74,9 @@ class Btn(v.Btn, SepalWidget):
                     DeprecationWarning,
                 )
 
-        # create the default v_icon
-        self.v_icon = v.Icon(children=[""])
+        # create the default v_icon (hidden)
+        self.v_icon = v.Icon(children=[""], left=True)
+        self.v_icon.hide()
 
         # set the default parameters
         kwargs.setdefault("color", "primary")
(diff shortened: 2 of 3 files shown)


dfguerrerom:
I don't see where you did this fix, main branch has already this two way biding, isn't it?

12rambau:
you are right I think I messed up with my branches. To make it useful I added the patch to #746

dfguerrerom:
Of course, the dictionary was being shared among all the different instances, great to have an exaplnation.

dfguerrerom (approved):

12rambau:
no problem, we both validated the layout_kwargs PR, both responsible

PR #742, opened 2023-02-07
fix: update package discovery

### LINKED ISSUE #734: many building deprecation issues in the ast build
https://github.com/12rambau/sepal_ui/actions/runs/4108942589/jobs/7090204215

cannot list them all but it needs to be handle before they get ignored by setuptool

12rambau:
Fix #734
need to be tested in the next fix release

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,20 +100,10 @@ sepal_ipyvuetify = "sepal_ui.bin.sepal_ipyvuetify:main"
 
 [tool.setuptools]
 include-package-data = true
-packages = [
-    "sepal_ui",
-    "sepal_ui.scripts",
-    "sepal_ui.frontend",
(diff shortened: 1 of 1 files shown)


dfguerrerom (approved):
cool

12rambau:
This issue made me realize something more critical, the latest version (2.15.0) is not including the data to the packages (.json, .parquet... etc) so it's barely working if you install it from pypi. I think my last commit is solving it but there is still "docs" and "tests" that are added to the distrib file (unwanted behavior). I've lost my whole afternoon on this one let's see if SO is better than me.

12rambau:
that was working since this morning (5h ago) I forgot to clean the cache before running the build command, it was always using an old file list... I'll bump a new version real quick to have the data folders back in the distrib

PR #743, opened 2023-02-07
refactor value inspector 

### LINKED ISSUE #731: improve vinspector rendering
In the current implementation everything is displayed at once, making it barely crawlable if many layers are displayed at the same time (which can be up to 67 in se.plan.

It would be nice to use a dense https://vuetifyjs.com/en/components/treeview/#usage

12rambau:
Fix #731 

- Change the inspector rendering using a `tree_view`. It makes it easier to read data with many different bands/layers
- Change the name: We are more or less following a convention for the controls of the map: XXXControl. ValueInspector was the only one not respecting it. I renamed the class "InspectorControl" and created a "deprecated" one to avoid regression.
- complete messages: there was one key still hard coded in the .py file  

Enregistrement de l’écran 2023-02-07 à 13 57 22

### DIFF
--- a/sepal_ui/mapping/__init__.py
+++ b/sepal_ui/mapping/__init__.py
@@ -19,6 +19,7 @@ Example:
 from .aoi_control import *
 from .draw_control import *
 from .fullscreen_control import *
+from .inspector_control import *
 from .layer import *
 from .layer_state_control import *
 from .layers_control import *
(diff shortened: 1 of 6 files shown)


dfguerrerom:
can we expand the first layer info by default once the map is clicked? so users will know that there's something inside?

12rambau:
> can we expand the first layer info by default once the map is clicked? so users will know that there's something inside?

done. I created a special parameter (`open_tree`) set to true by default to control this behavior as I already have applications in mind where I would like to prevent it

dfguerrerom (approved):

PR #756, opened 2023-02-10
add a fullscreen parameter to display bigger menu in map applications

### LINKED ISSUE #755: create fullscreen menucontrol
In some map application one would like to display a Menu covering everything (e.g. graphs or computed figures that are non geographical). It could be nice if a Menu could be sized to cover the whole surface of the map like this: 

Capture d’écran 2023-02-10 à 13 26 30

Main difficulties to overcome are: 
- be fullscreened even if the mapBtn is not the last one
- cover the full screen whatever the positioning of the mapBtn on the map
- work in both voila and notebooks (if it wasn't for notebooks, a simple "absolute stying would have done the trick")

12rambau:
Fix #755 

The idea is simle sometime I don't need the map anymore and I just want to display stuff to my users.
With this control option, I can create a control that fit nicely the frame of a map application (margin and icon width are absolute). It's only based on .css tricks so it cannot be coded outside of the lib. 

let me know if you want to tune it. here are 2 examples, one in an app, one with a standalone fullscreen map:

Enregistrement de l’écran 2023-02-10 à 17 18 32


12rambau:
The documentation bug is relative to this PR: https://github.com/sphinx-doc/sphinx/pull/10691
Nothing to do from here we just need for this fix to be merged.

dfguerrerom (commented):
I like the idea, it looks great, I've tested it and works as expected.
What do you think if we add a centered button at the bottom of the card saying "close"? As the card renders over all the window, the user could lose the sense of where it was opened (because it is not at the corner of the map btn as the others)... or perhaps a cross icon at the top right corner?
Apart from that I like the implementation.

12rambau:
> As I understand the menu card, I think that the content will be used to display text, right? what about centering the title?

I was more picturing a dashboard tile when you don't need the map anymore, let me send you a mockup as an example

> What do you think if we add a centered button at the bottom of the card saying "close"? As the card renders over all the window, the user could lose the sense of where it was opened (because it is not at the corner of the map btn as the others)... or perhaps a cross icon at the top right corner?

let me ask the question on UX stackexchange. I would like to avoid the big button but I'm not sure people will notice little "x" in the top header. there is still the "escape" otpion as well

12rambau:
comming back from UX, they suggested to fix the state "active" of the menu btn while the window is opened so that you can recognize where you're coming from. 

The x was considered but seems weird if not set everywhere. I'll go with that, update the tests and merge it

dfguerrerom (approved):

PR #758, opened 2023-02-13
fix: add a today method to datepickers

### LINKED ISSUE #752: add a today() method for the datepicker
It's something I do a lot, setting up the datepicker to today as: 

```python 

from sepal_ui import sepawidgets as sw 
from datetime import datetime 

dp = sw.Datepicker() 

# do stulff and as a fallback do 
dp.v_model = datetime.today().strftime("%Y-%m-%d")
```

Instead I would love to have something like: 
```python
dp.today()
```

what do you think ?

12rambau:
Fix #752

you can now create a dateipker set to today and/or dinamically change its date after initilization: 

```python 

sw.DatePicker().today()

# or 
dt = sw.DatePicker()
dt.today()
```

### DIFF
--- a/sepal_ui/sepalwidgets/inputs.py
+++ b/sepal_ui/sepalwidgets/inputs.py
@@ -156,6 +156,12 @@ class DatePicker(v.Layout, SepalWidget):
 
         return
 
+    def today(self) -> Self:
+        """Update the date to the current day."""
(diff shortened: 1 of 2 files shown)


12rambau:
it's here for a while and globally harmless, I'll merge it and make updates if required in following-up PRs

PR #774, opened 2023-03-01
fix: add root parameters in FileInput

### LINKED ISSUE #769: Restrict maximum parent level from InputFile
I have some apps where I’m interested on only search up to certain level, i.e., module_downloads, and I think that in the most of them, the user doesn’t need to go upper from sepal_user, once they start clicking, they could easily get lost over multiple folders.

what if we implement a parameter called: max_depth? We could use int values to this parameter, what do you think?

12rambau:
Fix #769 

- FileInput now embed a `root` traits that is piloting how high you can go up in the hierarchy. 
- It can be set at init or while computing. 
- if the current `v_model` is higher, reload will make the parent node disapear, but you'll still be able to go deeper

```python 
from pathlib import Path

fi = sw.FileInput(root=Path.home())
fi
```

### DIFF
--- a/sepal_ui/sepalwidgets/inputs.py
+++ b/sepal_ui/sepalwidgets/inputs.py
@@ -205,6 +205,9 @@ class FileInput(v.Flex, SepalWidget):
     clear: Optional[v.Btn] = None
     "clear btn to remove everything and set back to the ini folder"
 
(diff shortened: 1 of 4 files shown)


12rambau:
is_relative_to is not available in Python < 3.9

I'll add a monkey patch in utils

dfguerrerom (approved):

PR #785, opened 2023-03-08
fix: drop reclassify module

12rambau:
Fix #693 

I dropped everything at one, as it's never used in our applications. As there was lot of work involved, I created a specific branch with the content of the module: https://github.com/12rambau/sepal_ui/tree/reclassify_archive


12rambau:
as already discussed and considering the current heavy lifting work performed to optimize the tests, I'll drop it now. I can still be revived from the archive branch if necessary.

PR #796, opened 2023-03-18
fix: add key to any layer created with sepal-ui

### LINKED ISSUE #593: find layer by unequivocal key in a map
**Is your feature request related to a problem? Please describe.**
I'm developing a module and using the translator to have a 100% translatable app, the thing is that even though I can use a translatable name as a name to a layer, this is prone to errors.

**Describe the solution you'd like**
I would like to be able to use both `name` and a display `label` in the `add_layer` method, so the name can be used unequivocally as a key to performing search operations and the label will be used only to visualization purposes.

what do you think? @12rambau ?

12rambau:
Fix #593

Attempt to add an unequivocal key to every layer added to a SepalMap. 

In short I did 2 things:
- add a key parameter to the `add__layer` method. It's optional so anything that was already working should continue to work. If the key is not set, we normalize the name and use it instead (my guess is that if you're not specifying it, you are in fact not using it). 
- in `find_layer` if the key is a string, we first check the keys and then if nothing is found we test the names. My guess here is that no key will be the same as a name and if it's the case it's the same layer (i.g. we used the normalized version of the name and it didn't required any modifications).

I only added tests for `find_layer`, I will add a specific open if you like this modification. 

What do you think ?


tests/test_mapping/test_SepalMap.py
```diff
@@ -484,10 +486,14 @@ def test_find_layer(ee_map_with_layers: sm.SepalMap) -> None:
     with pytest.raises(ValueError):
         res = m.find_layer(50)
 
-    # search by layer
+    # search by layer name
     res = m.find_layer(m.layers[3])
     assert res.name == "NDWI harmonics"
 
+    # search by layer name
```
12rambau, line 493:
```suggestion
    # search by layer key
```

dfguerrerom:
I was too late on this, but it's okay to me, but I'm not sure if you still want to use different types for `key` param

12rambau:
@dfguerrerom never too late no worries, I've made the last modifications and it's now ready for review

dfguerrerom (approved):

PR #835, opened 2023-06-06
feat:  use localtileserver as native deps

12rambau:
people from kitware and I we finally finished the full rasterio implementation of lagre_image (https://github.com/girder/large_image/pull/1115) which allowed localtileserver to now only depends on rasterio (https://github.com/banesullivan/localtileserver/releases/tag/0.7.0). That's a very good news as it means local images can finally be shown in interactive maps within sepal-ui applications. 

In this PR: 
- I used localtileserver as a default dependency (remove any traces of lazy imports)
- I edited a bit the docstring for special cases (SEPAL being one of them) 

for @dfguerrerom: 
As SEPAL fall in the "distant jupyter" know dificulties (https://localtileserver.banesullivan.com/installation/remote-jupyter.html), I also imported a proxy generator in the lib. If I'm not wrong you should be able to make everything work by setting the `LOCALTILESERVER_CLIENT_PREFIX` env variable to "/api/sandbox/jupyter/proxy/{port}". I think you will need to work with @cdanielw to add this env variable in every user instances (e.g. in the default .bashrc) so that user get something on their screen without the need to go to their terminal.

Let me know if you think I should modify anything, I'm planning on merging it next week.


dfguerrerom:
Thanks for that @12rambau. I have tested it on SEPAL and it works smoothly.
Is it normal that the raster is lost when you zoom more than 12? On my end it totally disappears, I have tried it using two different images in two different envs.

12rambau:
That's unexpected, is there a `max_zoom` member to the created layer ?

dfguerrerom:
Localtilelayer sets `max_zoom` as the raster `default_zoom`, but when it comes to satellite imagery, users prefer to zoom at the pixel level. I have hardcoded 20 as `max_zoom` level taking into account that the max zoom of google satellite layers ranges from 18-21, what do you think?

12rambau:
Thanks for finding out, I would go directly to the max value of leaflet basemaps (24) let me update the value and fix the test and I'll request your review again

12rambau:
It is also forcing me to drop support of Python 3.7 as it's not supported by large_image. that's ok as py3.7 is reaching EOL on 2023-06-27

dfguerrerom (approved):

Issue #838, opened 2023-06-20
resize trigger doesn't work an potentially additional JS events


dfguerrerom:
resize event doesn't work even after https://github.com/12rambau/sepal_ui/pull/802 drawers were fixed

to reproduce:

- use the latest (v_2.16.4) version.
- open the panel app and add a new drawer with a map tile
- check that event resize doesn't work when you go to the drawer tab

PR #839, opened 2023-06-20
fix: closes 838. By using wrapper comments

dfguerrerom:
The problem here (after many wasted hours) was that we're reading the `js` files as text and so, apparently, the inline comments had a conflict with the `js` script, I have wrapped the comments with `/* ...*/` and apparently it works now.


12rambau:
fix https://github.com/12rambau/sepal_ui/issues/838

dfguerrerom:
@12rambau, I have closed this PR because checks didn't start, I'm trying to check why.

12rambau:
my bad, I was in auto pilot, I saw "closed" I thought it was "merged"

PR #845, opened 2023-07-25
build: finally drop setup.py

12rambau:
It is no longer needed since python 3.7


12rambau:
@dfguerrerom simply ignore 3.11 for the time being I'll work on it later

dfguerrerom (approved):

PR #853, opened 2023-08-03
fix: update planet mosaic list in tests

### LINKED ISSUE #850: the planet test are failing on regular basis
it seems the metadata of the downloaded images change on regular basis. I should update the tests so that they succeed in a consistent way

12rambau:
Fix #850


12rambau:
@dfguerrerom merging without review as it's going to avoid all other builds to fail

PR #856, opened 2023-08-03
fix: add `self` to `catch_error` decorator

### LINKED ISSUE #851: make alert optional in catch_error
In the catch_error decorator, the alert item is compulsory I suggest we do as for loading_button, i.e. make it optional and use the component set in the object by default: 
self.alert

### LINKED ISSUE #852: Use Optional instead of Union[None, ...]
Now that we don't support 3.7 any more, we should drop `typing_extentions` and stop using the Union[none, ...] pattern. 

2 option are possible: 
1.   `none|...`
2.   `Optional[...]`

I personnaly prefer 2.

12rambau:
Fix #852, Fix #851 

- fix the inconsistency between `catch_error` and `loading_button` by adding a `self` parameter to it. No need to specify the alert if the object have an `alert` member. 
- some typing refactoring

### DIFF
--- a/sepal_ui/scripts/decorator.py
+++ b/sepal_ui/scripts/decorator.py
@@ -199,7 +199,7 @@ def loading_button(
                         [custom_showwarning(w) for w in w_list]
 
             except Exception as e:
-                alert_.add_msg(f"{e}", "error")
+                alert_.add_msg(f"{e}", type_="error")
                 if debug:
                     button_.toggle_loading()  # Stop loading button if there is an error
                     raise e
--- a/sepal_ui/scripts/decorator.py
+++ b/sepal_ui/scripts/decorator.py
@@ -13,7 +13,7 @@ import warnings
 from functools import wraps
 from itertools import product
 from pathlib import Path
-from typing import Any, Callable, List, Optional, Union
+from typing import Any, Callable, List, Optional
 
 import ee
 import httplib2
@@ -125,8 +125,8 @@ def need_ee(func: Callable) -> Any:
 
 @versionadded(version="3.0", reason="moved from utils to a dedicated module")
 def loading_button(
-    alert: Union[v.Alert, None] = None,
-    button: Union[v.Btn, None] = None,
+    alert: Optional[v.Alert] = None,
+    button: Optional[v.Btn] = None,
     debug: bool = False,
 ) -> Any:
     """Decorator to execute try/except sentence and toggle loading button object.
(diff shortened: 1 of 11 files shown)


sepal_ui/scripts/decorator.py
```diff
@@ -60,27 +60,33 @@ def init_ee() -> None:
...
         The return statement of the decorated method
     """
 
     def decorator_alert_error(func):
         @wraps(func)
-        def wrapper_alert_error(*args, **kwargs):
+        def wrapper_alert_error(self, *args, **kwargs):
+
+            # Change name of variable to assign it again in this scope
+            alert_ = self.alert if not alert else alert
```
dfguerrerom, line 81: what about if parent object has not defined an `alert` component, can we add an exception here?

tests/test_scripts/test_decorator.py
```diff
@@ -22,22 +22,38 @@ def test_init_ee() -> None:
...
     obj = Obj()
 
-    obj.func1()
+    # should return an alert error in the the self alert widget
+    obj.func0()
     assert obj.alert.type == "error"
+
+    # should return an alert in the external alert widget
+    obj.func1()
+    assert alert.type == "error"
```
dfguerrerom, line 54: alert type will remain `error` after the previous test, so this error is not properly tested. I think we should: create this object as fixture and rest both scenarios and/or add the `self.alert.type = "info"` to the reset method in the alert component.

12rambau: nice catch, I prefer to revert to info type on reset (I think it's a bug in the first place)

sepal_ui/scripts/decorator.py
12rambau: isn't it already the case ? if you don't set up a alert and the object don't have an alert object, you should get something like "xxx object hase no "alert" member"

dfguerrerom: yes and not, you'll get an error, but that error will be too vague, that would be something like... "`None type doesn't have reset method`".
You don't really know where the alert is required, and perhaps you'll decide to create either a custom `w_alert` or `alert` and add it as parameter to the decorator, I mean, this feature would be in someway hidden, you'll have to really know how this has been written... or perhaps document it?...

tests/test_scripts/test_decorator.py
12rambau: I think solved in https://github.com/12rambau/sepal_ui/pull/856/commits/5d86ad169502ae61460a1f4fe71b75142501ef71

sepal_ui/scripts/decorator.py
12rambau: I think solved in https://github.com/12rambau/sepal_ui/pull/856/commits/95770ae6be052700d161fab194548f7fa541d467

dfguerrerom (approved):

Issue #836, opened 2023-06-09
is precommit working?


dfguerrerom:
I noticed it after committing on `localtileserver` branch, pre-commit hooks did not run, that's not normal, right? I did a random change on a test_commit branch and it passed (https://github.com/12rambau/sepal_ui/commit/96b6fb425558103b153cf3b3b26961b0194474b8).

12rambau:
pre-commit are run in local only if they are installed and they are run as part of the unit.yml test only when PR to main. 

so if the question is about local env maybe run: 
```
pre-commit install
```

and if it's about the CD/CI, you need to make PR to check if it runs.

dfguerrerom:
yes, I hadn't installed it on my side. I thought and error came up when you push something without rules in the repo.

dfguerrerom:
something changed... before you wouldn't have been able to push if you hadn't run the `pre-commit` hooks... in my last PR I was able to push without having `prettier` installed and so that hook didn't work...

12rambau:
where are you pushing from ?

12rambau:
any update on this one ?

dfguerrerom:
I believe it was failing because I didn't have install `pre-commit` on my local branch, and test didn't run on github because of something, but now it seems to be okay.

PR #859, opened 2023-08-03
made random changes

dfguerrerom:


dfguerrerom:
@12rambau in order to close https://github.com/12rambau/sepal_ui/issues/836, do you know why some of the tests were skipped?

12rambau:
yep, the test are all skipped if the pre-commit one fail. The idea is to limit the burden on my GEE account when test will not be succecfull anyway. You can see the order matrix here: https://github.com/12rambau/sepal_ui/actions/runs/5748795736?pr=859

dfguerrerom:
> yep, the test are all skipped if the pre-commit one fail. The idea is to limit the burden on my GEE account when test will not be succecfull anyway. You can see the order matrix here: https://github.com/12rambau/sepal_ui/actions/runs/5748795736?pr=859

really nice feature <3

PR #861, opened 2023-08-06
build: create a devcontainer

12rambau:
this devcontainer will be used when creating a codespace from this repository. it ensure nox and pre-commit are always installed and pre-commit hooks are activated. 

No excuses for mistakes !


dfguerrerom (approved):
nice one

PR #868, opened 2023-08-14
fix: externalyse geometries management

### LINKED ISSUE #860: use pygaul and pygadm in aoi_model
I finally developped 2 external libs to manage GAUL and GADM dataset, to avoid duplicating maintenance work I would suggest to rely on them in the aoi_model

12rambau:
I needed to get geometries both with and without GEE for my work and realized I didn't wanted to use the whole sepal_ui suits. The method developped for getting geometries were well designed so I exported it in their own libs: [pygadm](https://github.com/12rambau/pygadm) and [pygaul](https://github.com/12rambau/pygaul).

In order to avoid duplicate work, and to reduce the size of sepal_ui I proposed in this PR to rely on these 2 extra libs to get GAUL and GADM data. 

- I updated aoi_model so that `_from_admin` use `get_items` method from both lib
- I updated the aoi_view to use `get_names` in the `AdminSelect` dropdown
- dropped the parquet dataset and reduce the size of the lib 
- PyGAUL and pyGADM relies on requests_cahce for managing their download so they are faster than the regular downnloads we were using.

side effects: 
- I corrected the bug with the planet data_regression. 
- There was a small bug with the aoi_view when used without map (reset was bugging because of the w_draw v_model)

I checked it manually and didn't changed any of the tests so normally if we did a decent job at making them this modification should be transparent in applications. 

Fix #860


12rambau:
I recently added a class implementation of both of them, let's use this one instead

12rambau:
@dfguerrerom any chance to get a review on this one or should I proceed with the merge ?

dfguerrerom:
> @dfguerrerom any chance to get a review on this one or should I proceed with the merge ?

give me a couple of days more please, these days have been a bit crazy.

12rambau:
No problem I was just checking you were still around

dfguerrerom (approved):
this is a nice implementation, I'm not sure about the name of the classes but that's something non related with this PR.

PR #871, opened 2023-09-01
fix: display a message at the top of the notebook while loading

12rambau:
The message is manually placed at the top of the ui notebook in both template.
it uses the `loading-app` ID 
it's removed from the screen when loading of the widgets ends. 

Question: 
As we place apps in an overlay do we even need to extra CSS ? It seems to me that the markdown will simply go under it isn't it ?

### DIFF
--- a/sepal_ui/frontend/css/custom.css
+++ b/sepal_ui/frontend/css/custom.css
@@ -104,3 +104,9 @@ nav.v-navigation-drawer {
   width: calc(100vw - 80px);
   height: calc(100vh - 20px);
 }
+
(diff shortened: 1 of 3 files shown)


dfguerrerom:
We'll need that css otherwise, the message will displace the components of the app.
FYI: we are working with a designer to create a sepal loading gif, so probably we'll add something more stylized in all `sepal-contrib` apps.

12rambau:
In the meantime that's a good enough solution for the generic applications. 

You can ignore the tests, they crashed because of planet

dfguerrerom (approved):

Issue #885, opened 2023-09-21
show app version somewhere in the app


dfguerrerom:
I'm not sure how to show it in the `map app` but in the regular app we can add the version number of the module (if the module contains a file called `CHANGELOG.md`) at the bottom of the `navigation bar` or in the `app bar`.
If user clicks over this element, we could display a card dialog showing the `changelog` of the app.
@12rambau what do you think ?

PR #886, opened 2023-09-25
Version card

dfguerrerom:
related to #885 

- This PR creates a version card element that will be appended to the bottom of the `NavigationDrawer`.
- It will assume there's a `__version__.py` file in the root folder containing the current version
- It will assume the latest version of the module is in `release` branch
- If there's a release, it will use the text generated there as initial element, if not, only the changelog will be displayed.

[screen-capture.webm](https://github.com/12rambau/sepal_ui/assets/12363250/f17c1864-a7cf-414c-a397-2999d0e2f63a)

## WIP:
- I need to create the tests...
- But in the meantime, what do you think about this?

### DIFF
--- a/sepal_ui/scripts/utils.py
+++ b/sepal_ui/scripts/utils.py
@@ -15,7 +15,7 @@ import ee
 import httplib2
 import ipyvuetify as v
 import requests
-import toml
+import tomli
 from anyascii import anyascii
 from deprecated.sphinx import deprecated, versionadded
 from matplotlib import colors as c
@@ -388,8 +388,8 @@ def get_app_version(repo_folder: Pathlike = Path.cwd()) -> str:
 
     # check if the file exist
     if pyproject_path.exists():
-        # read the file using toml
-        pyproject = toml.loads(pyproject_path.read_text())
+        # read the file using tomli
+        pyproject = tomli.loads(pyproject_path.read_text())
 
         # get the version
         return pyproject.get("project", {}).get("version", None)
--- a/sepal_ui/scripts/utils.py
+++ b/sepal_ui/scripts/utils.py
@@ -7,12 +7,13 @@ import re
 import string
 import warnings
 from pathlib import Path
-from typing import Any, Sequence, Union
+from typing import Any, List, Sequence, Tuple, Union
 from urllib.parse import urlparse
 
 import ee
 import httplib2
 import ipyvuetify as v
+import requests
 from anyascii import anyascii
 from deprecated.sphinx import deprecated, versionadded
 from matplotlib import colors as c
@@ -128,10 +129,8 @@ def init_ee() -> None:
     """
     # only do the initialization if the credential are missing
     if not ee.data._credentials:
-
         # if the credentials token is asved in the environment use it
         if "EARTHENGINE_TOKEN" in os.environ:
-
             # write the token to the appropriate folder
             ee_token = os.environ["EARTHENGINE_TOKEN"]
             credential_folder_path = Path.home() / ".config" / "earthengine"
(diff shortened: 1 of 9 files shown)


12rambau:
I like the idea of adding a version number but I would like not to use a __version__.py file. We already support the pyproject.toml file to upport [entry-point] key. I would suggest to use [[project.version]] for the version number. Advantage, it is supported by commitizen so it will be possible to update version with a simple cz bump.

sepal_ui/scripts/utils.py
```diff
@@ -399,15 +400,12 @@ def get_app_version(github_url: str, branch="release") -> str:
     """
     repo_owner, repo_name = parse_github_url(github_url)
 
-    version_url = f"https://raw.githubusercontent.com/{repo_owner}/{repo_name}/{branch}/__version__.py"
-
+    version_url = f"https://raw.githubusercontent.com/{repo_owner}/{repo_name}/{branch}/pyproject.toml"
```
12rambau, line 403: I would not get the version number from the repository but from the application folder itself. Normally the folder is pulled from a repository and as such embed the .toml file. It ensures that the version displayed is consistent with what is installed on your machine and not what is available in distant repo. 
Also it avoids getting too much bound to github and start passively supporting other VCS like gitlab or azure devOps

dfguerrerom:
@12rambau two things here:
- I've fixed the coverage issues we had, and now I think we'll have an increase.
- I don't really know why the builds are failing with the `Translator`, I didn't touch it.
- the readthedocs documentation failed because of a missing `toml` package... I'm not sure how readthedocs creates the documentation, are we giving the `[doc]` requirements to that env?

12rambau:
> I'm not sure how readthedocs creates the documentation, are we giving the [doc] requirements to that env?

yes we are in this file:  https://github.com/12rambau/sepal_ui/blob/main/.readthedocs.yaml
adding the toml deps in [doc] extra dependencies should do the trick.

sepal_ui/scripts/utils.py
```diff
@@ -375,6 +374,94 @@ def check_input(input_: Any, msg: str = ms.utils.check_input.error) -> bool:
...
+        the version of the repository
+    """
+    # get the path to the pyproject.toml file
+    pyproject_path = repo_folder / "pyproject.toml"
+
+    # check if the file exist
+    if pyproject_path.exists():
+        # read the file using toml
+        with pyproject_path.open("r") as f:
+            pyproject = toml.load(f)
```
12rambau, line 390:
```suggestion
       pyproject = toml.loads(pyproject_path.read_text())
```

sepal_ui/sepalwidgets/app.py
```diff
@@ -443,6 +437,7 @@ def __init__(
         code: str = "",
         wiki: str = "",
         issue: str = "",
+        repo_folder: str = Path.cwd(),
```
12rambau, line 440: your hint expects a str and you serve a Pathlib.Path object. 

looking at the error in the code it seems it's also the source of the bug. Instead I would create somwhere (utils ?) a new Type: 
```python 
Pathlike = Union[str, pathlib.Path] 
```

and reuse it everytwhere we set file/dir names (to stay on the safe side). 

For the default value set it to `""` and fill it with Path.cwd() in the init method. At parsing stage, there is no cwd yet.

dfguerrerom: Done, but I didn't use the `Pathlike` everywhere... perhaps is better if we handle that in a different PR? we could create a `types.py` file (if that's something common)

12rambau (approved):
I'm happy with the latest change. As it's a rather old PR, I let you decide if you ant to add extra small things before merging.

PR #888, opened 2023-09-29
refactor: use tomli instead of toml

### LINKED ISSUE #887: use tomli instead of toml lib to parse toml file
It seems `toml` is not the default toml python parser anymore so I would suggest we move toward to most used one (the one used by pip itself): `tomli`

ref: https://dev.to/pypyr/comparison-of-python-toml-parser-libraries-595e

12rambau:
Fix #887

### DIFF
--- a/sepal_ui/bin/module_deploy.py
+++ b/sepal_ui/bin/module_deploy.py
@@ -18,6 +18,7 @@ import subprocess
 from pathlib import Path
 from typing import Union
 
+import tomli
 from colorama import Fore, Style, init
 
 import sepal_ui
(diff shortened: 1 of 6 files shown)


sepal_ui/bin/module_deploy.py
```diff
@@ -195,9 +196,9 @@ def main() -> None:
     print("Export the env configuration of your module...")
 
     # check that the local folder is a module folder
-    toml = Path.cwd() / "pyproject.toml"
+    Path.cwd() / "pyproject.toml"
```
12rambau, line 199:
```suggestion
```

Issue #893, opened 2023-10-05
navigation bar is not fixed


dfguerrerom:
I have already reported the issue on in the `ipyvuetify` repo, but I think we'll need to find a workaround to make our apps keep looking good until this issue is solved/reviewed by someone  https://github.com/widgetti/ipyvuetify/issues/280.

PR #894, opened 2023-10-05
App display bug

dfguerrerom:
- Aims to close https://github.com/12rambau/sepal_ui/issues/893


12rambau (changes requested):
I'm completely ok with the changes, I think the reference to the issue will be sufficient to remember why it's here. 

On the other hand I'm against living dummy content in the About_en.md file. It's overwritten by the module_factory so nobody will actually see it unless they pull the template directly. As we cannot perform test on the display, we will simply forgot why it's here. 

Could you revert the file to it's original state ?

dfguerrerom:
Done!

12rambau (approved):

PR #902, opened 2023-11-19
build(ipyleaflet): aims to close #901

### LINKED ISSUE #901: Ipyleaflet not compatible with jupyterlab 4
As I commented on https://github.com/openforis/sepal/issues/304, `ipyleaflet` events don't work with `Jupyterlab 4`. The problem was related to incompatibility of `js` libraries on the build, it was solved on https://github.com/cdanielw/ipyleaflet.
To speed up the implementation process I suggest to use that fork in this project.

dfguerrerom:
- I have changed the `ipyleaflet` requirements to https://github.com/cdanielw/ipyleaflet.git.
- I tried to use just the repo but I got an error on the `ipyleaflet` build in the documentation test since that environment doesn't have `yarn` which is required to compile and build the wheel, so I opted to use the compiled wheel that is already in the forked project.
- I have tested it on different environments and it works but Windows.

@12rambau let me know what do you think about this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
     "ipykernel",
     "ipyvuetify",
     "markdown",
-    "ipyleaflet @ git+https://github.com/cdanielw/ipyleaflet.git#egg=ipyleaflet",  # jupyter lab == 4 supported
+    "ipyleaflet @ https://github.com/cdanielw/ipyleaflet/releases/download/0.17.4/ipyleaflet-0.17.4-py3-none-any.whl",  # jupyter lab == 4 supported
     # earthengine management
     "earthengine-api",
     # to create a module
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
     "ipykernel",
     "ipyvuetify",
     "markdown",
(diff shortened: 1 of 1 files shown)


12rambau:
I cannot accept this proposal for several reason: 
- You cannot use wheel dependency in a Pypi distribution (it's there policy) 
- You cannot use a non published version in a conda build (It's even more strict than pipy) 
- I'm using the Windows version in my current job so it needs to be fully supported. 

what you could do for the sepal environment is force install the correct version like we do for the earthengine version. Just make sure to use a HUGE warning in the comments to not forget about it.

dfguerrerom:
> I cannot accept this proposal for several reason:
> 
> * You cannot use wheel dependency in a Pypi distribution (it's there policy)
> * You cannot use a non published version in a conda build (It's even more strict than pipy)
> * I'm using the Windows version in my current job so it needs to be fully supported.
> 
> what you could do for the sepal environment is force install the correct version like we do for the earthengine version. Just make sure to use a HUGE warning in the comments to not forget about it.

I understand.
I'm not sure if we can add `yarn` as dependency to the `readthedocs` build, but if that's possible that test might pass and the wheel would be created, I'm still have to test if that's possible in Windows as well. I'll keep you updated.

12rambau:
closed by the latest ipyleaflet merge

PR #903, opened 2023-11-30
Sepal pre release

dfguerrerom:
- I use this branch to update some of the features that are required but are not fully accpeted.

This PR it's just a draft, never will be an actual PR.

This PR can be deleted but NOT the branch


12rambau:
@dfguerrerom could you implement the modification you made for the authentication method in another PR ? 

I would like to start integrating your changes so you don't diverge too much from the main branch and I think it would be the first step to make sure the tests are running again on your machine.

PR #906, opened 2024-02-26
Update module_venv.py bin

dfguerrerom:
- It is just to update and better mimic the default behaviour as SEPAL creates the venv


dfguerrerom:
failing test are not related with the scope of the PR, they are happening because of the mandatory project authentication in gee.
that problem can be followed in https://github.com/12rambau/sepal_ui/pull/905

dfguerrerom:
closed in favor of https://github.com/openforis/sepal_ui/pull/926

Issue #909, opened 2024-03-29
make sepal_ui compatible with voila sepal-ui-template


dfguerrerom:
I've been working with a [voila template](https://github.com/dfguerrerom/voila-sepal-ui) to render sepal-ui based apps which has some cool features:
- Replace default jupyter spinning wheel and cell count number when loading
- Adds a kernel status icon
- Capture errors when it executes the cells and display them in a more user-friendly way
- Load app js scripts on a custom holder for scripts in the DOM (and here is where we need to add a tag to all vuetify template scripts)
- Minimize the DOM (diom).

It will also close https://github.com/12rambau/sepal_ui/issues/908 and https://github.com/12rambau/sepal_ui/issues/884

12rambau:
ping me again If I don't review this one before Monday

PR #910, opened 2024-03-29
feat: adapt to voila-sepal-ui template. #909

dfguerrerom:
This is related to https://github.com/12rambau/sepal_ui/issues/909. 
There are only few changes that I had to do. As I reported in https://github.com/widgetti/ipyvuetify/issues/308, `VuetifyTemplates` don't have a `metadata_id` and is not possible to identify them on a straighforward way, that's why I added a `class` tag to the main scripts we are using "copy to clipboard" and "fullscreen" for maps, in this way they will be displayed in the DOM when using `display` method.
I didn't set this tag to the other styles and js's we [have here](https://github.com/12rambau/sepal_ui/blob/68bfd2b89ada6e3422448f47a1ba8968f3442dec/sepal_ui/frontend/styles.py#L155, ) because they are obsolete within the new template.

update 04-10-24
=============

Actually, I simplified the full screen event by just adding/removing a css class, that will make the trick work and we'll use less code.


dfguerrerom:
@12rambau, I came up with a simpler solution for the full screen that doesn't require to use the vuetify template, and works well in both scenarios.

sepal_ui/mapping/fullscreen_control.py
```diff
@@ -97,7 +78,12 @@ def toggle_fullscreen(self, *args) -> None:
         # change button icon
         self.w_btn.children[0].children = [self.ICONS[self.zoomed]]
 
-        # zoom
-        self.template.send({"method": self.METHODS[self.zoomed], "args": []})
+        (
+            self.m.add_class("full-screen-map")
+            if self.zoomed
+            else self.m.remove_class("full-screen-map")
+        )
```
12rambau, line 85:
```suggestion
        if self.zoomed:
            self.m.add_class("full-screen-map")
        else:
            self.m.remove_class("full-screen-map")
```

I usually don't like it but it's currently shorter. I will request some extra method in ipywidget: https://github.com/jupyter-widgets/ipywidgets/issues/3905

12rambau (approved):
I let you decide if you want to accept my suggestion but for the rest it's good to go !

PR #913, opened 2024-05-03
feat: make theme change interactive

dfguerrerom:
This PR does the following:

- change default dark and light theme colors
- live theme colors change when theme button is clicked
- add new 'reactive' colors to the theme
- change the basemap tile to its light/dark variant if they are the default carto basemaps
- automatically update colors on disclaimers using the logic already set
- do some small test corrections 

See the result:

[screen-capture.webm](https://github.com/12rambau/sepal_ui/assets/12363250/c03a7d40-9558-4ab7-a968-4523c843655c)

@12rambau, the video doesn't show the transition so well but I think it is because of the number of frames that my recording tool use.

I know the test will fall because of the gee issue, but that PR stills on hold.


12rambau:
I love it, checked the code and found nothing to complain about. This feature is so much better than the previous implementation.

12rambau (approved):

PR #922, opened 2024-05-22
feat: Make a lazy loading of gee gdf..close #919.

### LINKED ISSUE #919: don't retrieve the geometries
Is there a need to "download" the AOI if the `AoiModel` is instantiated with the  `gee` parameter? 
If I create a GEE app, there are only few and specific cases (or is there even any?)  where I want to retrieve the features from GEE and then store them as `GeoDataFrame` in the object,
I think we should not make this expensive process as the default  behavior every time an AOI is created, this is not useful, and it is even worst if a vector file is selected from the client side: it uploads it and then it downlaods it.

dfguerrerom:
- Related to #919.
- Basically I create a lazy import on the `gdf` when it is requested using a full GEE AoiModel.
- I also changed the `AoiView` method to add both local and ee layers independently.

### DIFF
--- a/sepal_ui/aoi/aoi_model.py
+++ b/sepal_ui/aoi/aoi_model.py
@@ -503,9 +503,11 @@ class AoiModel(Model):
 
         if self.gee:
             coords = self.feature_collection.geometry().bounds().coordinates().get(0).getInfo()
-            return [coords[0][0], coords[0][1], coords[3][0], coords[3][1]]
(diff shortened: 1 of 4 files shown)


dfguerrerom:
@12rambau the other tests fail because of the issue with your planet key

12rambau:
find with me. 

I really like the smooth implementation with a property.

12rambau (approved):

PR #934, opened 2024-08-21
feat: draft async get assets

dfguerrerom:
from the current recursive function implementation the time to load 100 assets stored in multiple folders, the time was:

####################### Recursive #######################
Execution time: 13.889884233474731 seconds

By using the suggested implementation we could improve the result  by 80%. 

####################### Async Concurrent 2 #######################
Execution time: 2.823052406311035 seconds

### DIFF
--- a/sepal_ui/scripts/gee.py
+++ b/sepal_ui/scripts/gee.py
@@ -1,6 +1,7 @@
 """All the heleper methods to interface Google Earthengine with sepal-ui."""
 
 import asyncio
+import os
 import time
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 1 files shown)


dfguerrerom:
It works but I don't want to dispatch threads to process the requests, I would like to create a fully asyncio call...

dfguerrerom:
Adding max_concurrent_tasks parameter with a semaphore I think will be enough to be sure this feature won't cause any issue. I will merge this PR and later I will open a new one once https://github.com/dfguerrerom/ee-client/issues/3 integrates this changes.

PR #935, opened 2024-08-21
refactor: remove changelog

dfguerrerom:


12rambau (approved):

dfguerrerom:
@12rambau you installed the sphinx repo as dependency in `.readthedocs.yaml` file, which uses an unreleased version... I also used that one as the requirements in in the `pyproject.toml`, do you think it is fine if we just unpin those?

The problem with the build in readthedocs was because we were using the `mdinclude` directive to include the CHANGELOG.md, but if we are not updating it, I think we can just remove it, but, is there a reason for not `update_changelog_on_bump` ?

PR #956, opened 2024-11-15
Asset select

dfguerrerom:
When someone pastes an asset_id directly into the AssetSelect and then clicks a "button" which triggers an event, the AssetSelect validation doesn't get triggered until the button event finishes... this causes validations errors.

This PR adds an optional argument that:

- Triggers validation of AssetSelect with a debounce when the `update:search-input` event is emitted.

I don't foresee any issue with this change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,8 @@ dependencies = [
     "natsort",
     "typing-extensions",
     "nest-asyncio", # planet API interaction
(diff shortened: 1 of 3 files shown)


dfguerrerom:
I'm bypassing the checks and merging since the error is unrelated with this PR, I have opened an issue to track it: https://github.com/openforis/sepal_ui/issues/957

PR #1011, opened 2026-07-23
feat: browser-owned locale resolution (LocaleState, live switching, no config writes)

SerafiniJose:
## What

Makes locale handling work the way theme handling already does (`Theming.vue` +
`ThemeState`): resolved in the browser, held in session-scoped state, switched
live with no page reload.

- **New `LocaleState`** (`pysepal/solara/locale.py`), structurally mirroring
  `ThemeState`: a `Unicode` trait holding the *resolved* locale code, registered
  per kernel by `SessionManager`, with `resolve_locale_state()` /
  `use_locale()` helpers and a pure `match_offered_locale()` matcher.
- **`LocaleSelect` (the Solara `VuetifyTemplate` in `vue_app.py`) resolves the
  locale in the browser** on mount:
  `localStorage[":sepalUi:locale"]` → a one-time read-only seed from
  `~/.sepal-ui-config` (copied into localStorage when it wins, so the migration
  happens once) → `navigator.language` → `"en"`. Every candidate is validated
  against the app's offered locales: exact match → bare primary subtag if
  offered → first offered variant sharing the primary subtag → fall through.
- **The pick is pushed to Python by direct trait assignment**
  (`this.selected_locale = code`). The old template used
  `this.$emit("update:selected_locale", …)` — the Vue `.sync` convention, which
  needs a binding parent and is silently dropped when the widget is a root
  widget, so picks never persisted (the "resets to English after a refresh"
  bug). Direct assignment on a synced `VuetifyTemplate` trait is the mechanism
  the rest of pysepal's templates already use.
- **`LocaleSelect` no longer writes `~/.sepal-ui-config`.** Language changes are
  live; the "refresh the page" alert is gone.
- **`MapApp` binds the session `LocaleState`** to its language selector, exactly
  as it already binds `ThemeState` to its theme toggle (`_coerce_locale_select`
  mirrors `_coerce_theme_toggle`), and accepts `locale_state` / `translator`
  kwargs. Without this, the default `MapApp` language button would relabel
  itself from `navigator.language` while every string stayed in the old
  language.

## Why

In 3.7.0, `LocaleSelect` persists to `~/.sepal-ui-config` via `set_config`, and
`Translator(target="")` reads that file at construction. Three consequences:

1. **Any translated string an app resolves silently depends on the developer's
   machine-global config file.** Building a `Translator` with no explicit target
   (`Translator(folder)` — the common case) calls `find_target`, which reads
   `[sepal-ui] locale` from `~/.sepal-ui-config` and picks the target language
   from it. So the language a test sees is whatever locale the machine running
   it happens to have set — a single file in `$HOME`, shared across every
   sepal-ui app and rewritten the last time any of them called `set_config`. It
   is not an environment variable a test run can scope; a suite cannot be
   hermetic without a conftest fixture that patches `config_file`.
2. **A language change needs a page reload** — the translator is rebuilt only at
   kernel start.
3. **The `$emit`-based template never persisted picks at all** for a root
   widget.

Theme already solved all three browser-side. This PR applies the same model to
locale. Apps opt in by passing an explicit `target` (from `LocaleState`) to
`Translator`; **`Translator(target="")` keeps its legacy config-reading
behavior unchanged**, so existing apps are not forced to migrate.

## Compatibility

- `Translator(target="")` is untouched — same config-file read path as before.
- Existing apps that construct `LocaleSelect(translator=...)` without a
  `locale_state` still work; the widget simply has no state to mirror into.
- **Behavior change for apps that adopt `LocaleState`:** locale now follows the
  browser/localStorage, not `~/.sepal-ui-config`. The legacy config value is
  honored once (seeded and migrated to localStorage) and then never read again.
  Cross-app locale sync through the shared config file is intentionally dropped
  beyond that one-time seed — the same trade-off theme already made.

## Testing

- `tests/test_solara/test_locale.py` — the matching algorithm
  (exact/primary/variant/none, the `es`→`es-ES` migration case), `LocaleState`
  behavior, `resolve_locale_state` fallback determinism (a constant `"en"` that
  never reads config — this is what makes downstream test suites deterministic),
  and the per-session registration.
- `tests/test_sepalwidgets/test_LocaleSelect.py` — config seeding from a tmp
  file, bidirectional `LocaleState` binding, and that selecting a locale updates
  the state and **never** writes config (asserted by patching
  `pysepal.scripts.utils.set_config` and checking it is not called).
- `pysepal/solara/__init__.py` exports are covered; `pre-commit run
  --all-files` is clean (ruff import order, prettier on the `.vue`).

**Frontend coverage caveat:** pysepal has no JS test harness, and adding one is
out of scope here — the same coverage level as `Theming.vue`. The Vue
`matchOffered` is a line-for-line transcription of the unit-tested Python
`match_offered_locale`, and the browser-only behaviors (localStorage
persistence/migration/failure, `navigator.language` detection, and that a
remount does not clobber a live pick) were verified manually in a downstream
app.

## Known limitations / follow-ups (not addressed here, disclosed for the reviewer)

- **Offered locales are intersected with `pysepal/data/locale.parquet`**, which
  has regional codes (`es-ES`, `pt-BR`, …) but no bare `es`/`pt`. An app whose
  catalog uses a bare code that the parquet lacks — including pysepal's own
  bundled `message/es/` — cannot surface it in the picker, and a legacy
  `locale = es` seed is dropped rather than migrated. This is pre-existing
  filtering, but this PR makes it load-bearing (previously
  `Translator(target="")` bypassed the parquet). Proposed as a **separate
  issue** rather than widening this PR, since fixing it changes the picker's
  contents for every downstream app.
- **Theme still has the machine-global-config dependency that locale now
  sheds.** `Translator`/theme resolution via the config file is unchanged.
  Worth a follow-up so the two mechanisms match; deliberately not touched here.
- **Name collision:** `pysepal.sepalwidgets.LocaleSelect` (the exported
  `Menu`-based widget in `app.py`) and the reworked `VuetifyTemplate`
  `LocaleSelect` in `vue_app.py` now behave *oppositely* on persistence — the
  exported one still writes config, and its docstring still promises to. A
  rename would make the intent unambiguous; flagged for discussion on how you'd
  prefer to sequence it.


dfguerrerom:
Thanks for the PR @SerafiniJose. Can you please create an example about how to do this implementation in the example app in the template? `pysepal/templates/solara/solara_map_app/app.py` thank you.

dfguerrerom:
[generated with opus 5 (1M context), thinking level: max]

@SerafiniJose this work now lives on the pysepal 4.0 branch as #1028, so it ships with v4 instead of v3. Your browser-resolution commit is preserved there with your authorship.

Three things changed in the port:

- The `~/.sepal-ui-config` seed is gone. `pysepal.conf` does not exist in 4.0, so there is nothing left to migrate from.
- `LocaleState` moved to the v4 scope registry (`get_scoped_state`) instead of `SessionManager`, matching how `ThemeState` works in 4.0.
- `MapApp` takes locale codes instead of a `Translator`. `Translator` subclasses `dict` and reacton flattens it, so `translator=` breaks when the app uses `MapApp.element()`.

Two bugs turned up while testing it in a browser, both now fixed in #1028:

- `const STORAGE_KEY` was declared outside `export default`. ipyvue does not evaluate the file as a module, so the constant was undefined inside the methods. Both localStorage calls raised `ReferenceError` into their empty `catch` blocks, so the language choice was never saved and never read back.
- `MapApp` builds its `RightPanel` child once at construction and never updates it, so only the app title changed language.

The picker also now offers every catalog an app ships. It used to be filtered against `locale.parquet`, which has `es-ES` but no bare `es` — so pysepal's own `message/es/` and `message/ru-RU/` could never be selected.

Closing this PR in favour of #1028.

PR #1016, opened 2026-07-24
fix(mapping): only prune foreign GDAL_DATA/PROJ_LIB, keep own-prefix paths

SerafiniJose:
## Problem

`pysepal/mapping/sepal_map.py` unconditionally deletes `GDAL_DATA` and `PROJ_LIB` from `os.environ` at import time (an old workaround for the SEPAL system GDAL breaking the wheel-bundled rasterio).

In conda environments those variables point at the environment's **own** GDAL data directory, and deleting them poisons GDAL data discovery for the environment's own GDAL stack. The full failure chain, observed in a conda-based app (spatial_risk):

1. `import pysepal.mapping.sepal_map` deletes `GDAL_DATA` (set by conda activation to `$CONDA_PREFIX/share/gdal`).
2. The pip rasterio wheel, imported afterwards, registers its bundled `rasterio/gdal_data` path in its internal GDAL config (correct only for its private libgdal).
3. When a raster layer is displayed, localtileserver's `TileClient` forwards rasterio's env options **process-wide** into `os.environ` for its background tile-server thread — so `GDAL_DATA` now points at `site-packages/rasterio/gdal_data`, which contains no `header.dxf`.
4. A later lazy `import pyogrio` (e.g. `GeoDataFrame.to_file`) runs its `header.dxf` sanity check against the poisoned `GDAL_DATA` and raises:

   ```
   ValueError: Found GDAL data directory at .../envs/<env>/share/gdal but it
   does not appear to correctly contain GDAL data files
   ```

## Fix

Replace the unconditional deletion with `_prune_foreign_gdal_env()`: delete `GDAL_DATA`/`PROJ_LIB` only when they point **outside** `sys.prefix`.

- SEPAL case unchanged: a system GDAL exporting `/usr/share/gdal` into a venv is still pruned.
- Conda/venv case fixed: paths inside the interpreter's own prefix belong to the environment's own GDAL stack and are kept.

## Testing

- New unit test `test_prune_foreign_gdal_env` covering foreign-path pruning, own-prefix preservation, and unset variables.
- Manually verified in the affected conda env: `GDAL_DATA` survives the sepal_map import, `import pyogrio` succeeds, and `GDAL_DATA=/usr/share/gdal` is still removed.

### DIFF
--- a/pysepal/mapping/gdal_env.py
+++ b/pysepal/mapping/gdal_env.py
@@ -1,20 +1,47 @@
-"""Sanitize GDAL/PROJ data environment variables."""
+"""Sanitize GDAL/PROJ data environment variables.
+
+SEPAL's system GDAL exports GDAL_DATA, PROJ_LIB and PROJ_DATA into every module
+venv, pointing at data files for a different GDAL build than the venv's wheels.
+That breaks raster reads, so drop them before anything imports rasterio.
+
+Paths inside the interpreter's own prefix are kept: they are this environment's
+own GDAL stack, and pyogrio and osgeo break without them. This cannot fix a mixed
+conda/wheel install, where a wheel's own libgdal lives in that prefix too.
+"""
 
 import os
 import sys
+from pathlib import Path
+
+GDAL_ENV_VARS = ("GDAL_DATA", "PROJ_LIB", "PROJ_DATA")
+"""Data-directory variables to sanitize.
+
+``PROJ_LIB`` is PROJ's pre-9.1 name for ``PROJ_DATA``; both are still exported in
+the wild, so a foreign PROJ install can arrive through either one.
+"""
+
+
+def _is_own_path(path: str) -> bool:
+    """Whether ``path`` resolves under one of this interpreter's own prefixes.
+
+    ``sys.base_prefix`` differs from ``sys.prefix`` inside a venv, and a venv
+    layered on a conda env inherits that env's GDAL stack, so both count as ours.
+    """
+    if not path:
+        return False
+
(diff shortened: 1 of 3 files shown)


dfguerrerom:
[Generated with aid of claude-opus-5.max]

Thanks for tracking this down, the root cause is right. I pushed a few things on top rather than send it back:

- `PROJ_DATA` — PROJ ≥ 9.1's name for `PROJ_LIB`, so the PROJ half of this was a no-op on any modern conda env.
- Path-aware prefix check instead of `startswith` — a sibling env like `…/envs/pysepal-other` matched as "own".
- `sys.base_prefix` as well as `sys.prefix`, so a venv layered on conda keeps that env's GDAL stack.

Worth saying why this hit your env and not ours: you have rasterio/GDAL installed from pip rather than conda-forge. A pip rasterio wheel ships its own `rasterio/gdal_data`, so once we delete `GDAL_DATA` rasterio registers that private path and localtileserver forwards it process-wide into `os.environ`. In an all-conda env `search_wheel()` returns `None`, rasterio registers nothing, and there's nothing to forward — which is why none of us ever saw it. Mixing pip wheels with conda-forge for the geospatial stack will keep producing this class of breakage, so it's worth reinstalling rasterio/GDAL from conda-forge.

One detail worth knowing: the path in that pyogrio `ValueError` is always `sys.prefix/share/gdal`. It's the fallback `init_gdal_data()` tries *after* the poisoned value fails, not the poisoned value itself — so the traceback is telling us the env's own `share/gdal` also failed.

PR #1017, opened 2026-07-25
feat(solara/aoi): restore AoiView from a pre-populated value at mount

SerafiniJose:
## Problem

The Solara `AoiView` cannot be programmatically restored: passing a rebuilt `AoiResult` as `value` renders a blank picker (method select empty, admin cascade unseeded, asset field empty). Apps that persist AOI selections across sessions (project save/load) currently have to vendor the component to get restore-on-load. The legacy sepalwidgets `AoiModel` already supports this via its `vector`/`admin`/`asset` constructor defaults — this PR brings the Solara component to parity.

## Changes

- **AoiView**: when `value` holds an `AoiResult` at mount, seed the method select, admin cascade, draw name/control and asset field from it, then auto-confirm the selection (map layer + success feedback) reusing the loaded result instead of reprocessing. Guards cover reacton's double effect-run and value churn during load-triggered remounts.
- **AdminLevelSelector**: new `initial` seed restores the full cascade from the final admin code via `admin_parent_chain()` (GAUL parquet lookup in `aoi/admin.py`).
- **AssetSelectComponent**: new `initial` seed re-applies a persisted `{asset_id, type, column, value}` selection (asset id validated and columns loaded as if picked, column/value seeded once the columns arrive).
- **AoiResult**: new `asset` field round-trips the asset picker inputs (populated by `process_asset`) so ASSET selections can be persisted and restored without a side-channel. Appended last with a default — fully backward compatible.

All new parameters are optional; existing apps are unaffected. The legacy `pysepal.aoi` (sepalwidgets) stack is untouched.

## Tests

Hermetic — no network or EE access (pygaul lookups and admin fetches mocked, fake GEE interface for the asset flow):

- `test_aoi_view_restore.py` — restore across every method (ADMIN0/1/2, SHAPE, POINTS, DRAW, ASSET): method select seeded, auto-confirm reached, loaded value preserved. DRAW additionally re-seeds the DrawControl (restored AOI stays editable) and re-draws the map layer; ADMIN1 restores both cascade levels; ASSET seeds the combobox from `AoiResult.asset`.
- `test_admin_selector_restore.py` — regression: the ADMIN1/ADMIN2 cascade must survive `initial` being bound to the live reactive the selector drives (whose mount-time reset wiped the chain) and reacton's double effect-run.
- `test_admin_parent_chain.py` — parent-chain lookup unit tests (mocked `pygaul._df`), including unknown/falsy codes and lookup failure returning `{}`.
- `test_aoi_asset.py` — `process_asset` attaches the selection inputs; `AoiResult.asset` defaults to `None`.

`tests/test_solara`: 188 passed on this branch. Legacy `tests/test_aoi`: unaffected, passing.

### DIFF
--- a/pysepal/solara/components/aoi/admin.py
+++ b/pysepal/solara/components/aoi/admin.py
@@ -184,6 +184,37 @@ def fetch_admin_items(
     return item_list
 
 
+def admin_parent_chain(method: str, code: Optional[str]) -> Dict[int, str]:
+    """Map a final GAUL admin code to its full ``{level: code}`` parent chain.
+
+    Restoring an ADMIN1/ADMIN2 cascade needs the parent codes (to load the
+    child dropdown lists), but only the final code is usually persisted. Look
+    the parents up in pygaul's GAUL 2024 parquet. Returns ``{}`` when the code
+    is falsy/unknown or the lookup fails, so callers can fall back to
+    restoring the method only.
+
+    Args:
+        method: The admin method ("ADMIN0", "ADMIN1", or "ADMIN2")
+        code: The final admin code of the persisted selection
+
+    Returns:
+        ``{level: code}`` for every level from 0 down to the method's level,
+        or ``{}`` when the chain cannot be resolved.
+    """
+    if not code:
+        return {}
+    try:
+        level = int(method[-1])
+        df = pygaul._df()
+        rows = df[df[f"gaul{level}_code"].astype(str) == str(code)]
+        if rows.empty:
+            return {}
+        row = rows.iloc[0]
+        return {lvl: str(row[f"gaul{lvl}_code"]) for lvl in range(level + 1)}
+    except Exception:
+        return {}
+
+
 async def process_admin(
     method: str,
     admin_code: str,
--- a/pysepal/solara/components/aoi/admin.py
+++ b/pysepal/solara/components/aoi/admin.py
@@ -4,6 +4,7 @@ Functions for selecting and processing administrative boundaries using
 FAO GAUL 2024 data (both WFS for non-GEE and sat-io asset for GEE).
 """
 
+import asyncio
 import json
 from pathlib import Path
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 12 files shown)


dfguerrerom:
[Written with the aid of Claude Opus 5 (`claude-opus-5[1m]`), at maximum reasoning effort.]

I close this PR, and a new version replaces it. But the problem in the description is correct, and it caused the new design. Thank you for it. There are two reasons to write the code again, and not to rebase it.

The first reason is `AoiResult`. It cannot be the input for a restore. It holds a GeoDataFrame and an `ee` object, so an application cannot save it to disk. It also does not keep the picker inputs for the SHAPE, the POINTS and the ASSET methods. A new `AoiSpec` class does this task now. `AoiSpec` is small, frozen and safe for JSON,and it holds the inputs of all the seven methods. A new `spec`/`on_spec` channel on `AoiView` carries it in two directions. The `value`/`on_value` channel does not change, and it continues to carry the `AoiResult`. The legacy `AoiModel` already did this with its `admin`, `vector_json`, `point_json`, `geo_json` and `asset_json` traitlets. The Solara version removed the function.

The second reason is the code that sets the initial values: `initial=`, `restored_method`, `restore_pending`, `_seeded`, `_UNSET` and the two parent refs. This code corrects a problem in the components. Solara does not have an `initial` parameter. In Solara, `value` sets the initial state, and a component calculates its internal buffers from `value` again when `value` changes. Also, the loaders in `AdminLevelSelector` do two tasks: they get the items, and they clear the code of the child level. Because of this, the restore operation and the loaders write to the same state. The new code keeps the cascade in one tuple of codes, and the loaders do not write a code. The spec keeps all the GAUL codes. So `admin_parent_chain()` and its scan of the private `pygaul._df()` table are not necessary.

I keep two changes from this PR. The unmount correction is a different problem, and it causes a loss of data. `use_reactive` returns the reactive of the caller. So the unmount operation made the AOI of the application `None`. This correction goes in first, as its own commit. The change to the event loop goes in a different PR. It is the opposite of the rule in `docs/guides/ solara-gee-patterns.md`, and `main` has new code in `gee_interface.py`. So it needs its own test against the v4 interface.

The new branch is `feat/aoi-spec-restore`.

